PR #8, opened 2019-10-31
using TAB key event

bresilla:
a key event that makes use of TAB key to cycle through menus,
radiobuttons and user input


SanJ2312:
Shouldn't it be done for other components like input too ?

.gitignore
```diff
@@ -0,0 +1,8 @@
+/build/
+/bin/
+/lib/
+
+.cs_files
+.notifier
+/.undodir/
+.gdb_history
```
ArthurSonzogni: Could you please remove this?

src/ftxui/component/container.cpp
```diff
@@ -61,6 +61,15 @@ bool Container::VerticalEvent(Event event) {
     }
   }
 
+  if (event == Event::Tab) {
+    if (selected_ != int(children_.size()) - 1) {
+      selected_++;
+      return true;
+    } else {
+        selected_ = 0;
+    }
```
ArthurSonzogni: Suggestions:
~~~cpp
selected = (selected + 1) % children.size();
return true;
~~~

I think you should return true in both cases. Otherwise, it would cause issues when you have nested containers

src/ftxui/component/menu.cpp
```diff
@@ -27,6 +27,11 @@ bool Menu::OnEvent(Event event) {
     new_selected--;
   if (event == Event::ArrowDown || event == Event::Character('j'))
     new_selected++;
+  if (event == Event::Tab){
+    new_selected++;
+    if (new_selected > (int)(entries.size()-1))
+        new_selected = 0;
```
ArthurSonzogni: Suggestions:
~~~cpp
new_selected = (new_selected + 1) % entries.size();
~~~

src/ftxui/component/radiobox.cpp
```diff
@@ -28,6 +28,11 @@ bool RadioBox::OnEvent(Event event) {
     new_focused--;
   if (event == Event::ArrowDown || event == Event::Character('j'))
     new_focused++;
+  if (event == Event::Tab){
+    new_focused++;
+    if (new_focused > (int)(entries.size()-1))
+        new_focused = 0;
```
ArthurSonzogni: Suggestion:
~~~
new_focused = (new_focused + 1) % entries.size();
~~~

ArthurSonzogni:
> Shouldn't it be done for other components like input too ?

I don't think so. The input component is not a container.

ArthurSonzogni:
No need to update anything. I will directly cherry-pick your CL and append my suggestions. I found a few glitch locally. I would like to fix them.

ArthurSonzogni:
I will also add the reverse-tab.

ArthurSonzogni:
I updated your pull request. I added support for shift + tab. I also refactored a bit this area.
Please let me know if the new patch looks good to you.

bresilla:
> I updated your pull request. I added support for shift + tab. I also refactored a bit this area.
> Please let me know if the new patch looks good to you.

Perfect.
Thank you.
I am using FTXUI in a project of mine and it seems very cool for now. Maybe in the future some more documentation would be better for new-comers. If i get comfortable with it, i can help add some documentation on this project and submit another PR.

Thanks again :)

ArthurSonzogni:
Any PR are welcome!
I am kind of ignorant about whether or not a lot of people are using FTXUI.
If you have any suggestions about FTXUI, please let me know.

PR #9, opened 2020-02-02
Fix compilation issue on macOS(CLang compiler) 

alfishe:
Fixes compilation error for macOS Clang compiler.

> '/src/ftxui/component/screen_interactive.cpp:36:5: error: use of undeclared identifier 'quick_exit'

For some reason macOS cstdlib doesn't contain definition for quick_exit()

### DIFF
--- a/src/ftxui/component/screen_interactive.cpp
+++ b/src/ftxui/component/screen_interactive.cpp
@@ -11,6 +11,11 @@
 #include "ftxui/screen/string.hpp"
 #include "ftxui/screen/terminal.hpp"
 
+#if defined(__clang__) && defined (__APPLE__)
+    // Quick exit is missing in standard CLang headers
+    #define quick_exit(a) exit(a)
+#endif
+
 namespace ftxui {
 
 static const char* HIDE_CURSOR = "\e[?25l";


ArthurSonzogni:
Thanks! You are probably the first one to use it on macOS.

I think this is because std::quick_exist must be used "with the std::" namespace. Does it fix the issue?

alfishe:
On my macOS 10.14 there is no _LIBCPP_HAS_QUICK_EXIT defined for CLang C++ 11. Probably Apple released it in newer version for 10.15

So additional ifdef check could be added before rollback to exit(). But it's so minor thing

alfishe:
> /src/ftxui/component/screen_interactive.cpp:36:10: error: 'quick_exit' is not a member of 'std'
>    36 |     std::quick_exit(SIGINT);
>       |          ^~~~~~~~~~

even with std::quick_exit prefixing and check for _LIBCPP_HAS_QUICK_EXIT is defined. So immediate rollback seems to be a best option

ArthurSonzogni:
Thanks for checking with "std::"
This patch looks good to me. Thanks!

PR #10, opened 2020-02-11
CMakeLists.txt rewrite

GiuseppeCesarano:
Features:

- ~~Automatic indexing of source files for each library~~
- Exported targets for easy use as git sub-module or with projects like [CPM.cmake](https://github.com/TheLartians/CPM.cmake)
- Options to exclude unwanted build targets

I tested this file and it is backwards compatible with the old one.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,10 @@ project(ftxui
   VERSION 0.1
 )
 
+option(FTXUI_BUILD_EXAMPLES "Set to ON to build examples" ON)
+option(FTXUI_BUILD_TESTS "Set to ON to build tests" ON)
+option(FTXUI_ENABLE_INSTALL "Generate the install target" ON)
+
 enable_testing()
 
 cmake_minimum_required(VERSION 3.0)
@@ -56,6 +60,10 @@ add_library(component
   src/ftxui/component/toggle.cpp
 )
 
+add_library(ftxui::screen ALIAS screen)
+add_library(ftxui::dom ALIAS dom)
+add_library(ftxui::component ALIAS component)
+
 target_link_libraries(dom PUBLIC screen)
 target_link_libraries(component PUBLIC dom)
 target_link_libraries(component PUBLIC Threads::Threads)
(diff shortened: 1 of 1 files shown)


examples/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
-add_subdirectory(component)
-add_subdirectory(dom)
-add_subdirectory(util)
+if(FTXUI_BUILD_EXAMPLES)
```
ArthurSonzogni: What about moving this in the main CMakeLists.txt. Wrapping "add_subdirectory(examples")

GiuseppeCesarano: Clearly a better solution if you plan to use the examples/CMakeLists.txt only for build the examples. If that's the case I'm more than happy to make the changes.

ArthurSonzogni:
Thanks!

> **Automatic indexing of source files for each library**
I would prefer not to do this. Mainly for the reasons here:
https://stackoverflow.com/questions/1027247/is-it-better-to-specify-source-files-with-glob-or-each-file-individually-in-cmak/18538444

CMakeLists.txt
```diff
@@ -1,121 +1,86 @@
...
-)
+  VERSION 0.1)
 
 enable_testing()
 
 cmake_minimum_required(VERSION 3.0)
+
+option(FTXUI_BUILD_EXAMPLES "Set to ON to build examples" ON)
+option(FTXUI_BUILD_TESTS "Set to ON to build tests" ON)
+option(FTXUI_ENABLE_INSTALL "Generate the install target" ON)
```
ArthurSonzogni: Could you explain why we need the "FTXUI_ENABLE_INSTALL" option?
Why would the user set this to OFF?

examples/CMakeLists.txt
ArthurSonzogni: I was thinking about doing:
~~~cmake
if (FTXUI_BUILD_EXAMPLES)
  add_subdirectory(examples)
endif()
~~~

I am not sure to understand fully what you meant in you previous response.

CMakeLists.txt
GiuseppeCesarano: If the user wants to use the library as a dependency for a single program and does not want to install it globally in the system, the ability to remove the install target slightly speeds up the build

examples/CMakeLists.txt
GiuseppeCesarano: In this way the file in examples is not included, if its sole purpose is to build the examples it is better to use the approach proposed by you, if instead the file will have other functions it is better that the controls are in the file itself. So it depends on what are your intentions with the file

ArthurSonzogni: I see!
The goal is to build the examples only. So I would prefer adding the "if" in the around the add_subdirectory.

CMakeLists.txt
ArthurSonzogni:
```
and does not want to install it globally in the system
```
I don't understand how the user who doesn't want to install FTXUi would end up install it?
You are probably right, but I don't understand.

```
the ability to remove the install target slightly speeds up the build
```
I am not convinced this can be humanly perceivable.

CMakeLists.txt
```diff
@@ -53,69 +56,68 @@ add_library(component
...
-
-  target_include_directories(dom_tests
-    PRIVATE ${GTest_INCLUDE_DIRS}
-  )
-  gtest_discover_tests(dom_tests)
-  add_test(dom_tests dom_tests)
-
+# https://stackoverflow.com/questions/24295876/cmake-cannot-find-a-googletest-
+# required-library
+if(FTXUI_BUILD_TESTS)
```
ArthurSonzogni: Nit: I would move this inside the "if" line 104. This will remove one indentation level.

CMakeLists.txt
```diff
@@ -1,23 +1,26 @@
...
-add_library(screen
-  src/ftxui/screen/box.cpp
-  src/ftxui/screen/screen.cpp
-  src/ftxui/screen/string.cpp
-  src/ftxui/screen/terminal.cpp
-  src/ftxui/screen/wcwidth.cpp
-)
+add_library(
+  screen
+  src/ftxui/screen/box.cpp src/ftxui/screen/screen.cpp
```
ArthurSonzogni: Nit: Please keep one file per line, every line sorted alphabetically.

GiuseppeCesarano:
Thanks for the tips, I cleaned the code as requested!

ArthurSonzogni:
Perfect!

ArthurSonzogni:
Thanks for applying all those comments. LGTM!

PR #32, opened 2020-07-23
Fixed P conversion

GiuseppeCesarano:
Simply removing the 'P' case the conversion works fine. I didn't get the intention behind this case so feel free to reject the pull request if this fix isn't suitable.


ArthurSonzogni:
The bug has been introduced in:
https://github.com/ArthurSonzogni/FTXUI/commit/001a0ae92579da947cc5d8fc2fb1143aa5336cc9

This is very weird didn't found this before.
Thanks for your patch!

Issue #37, opened 2020-08-02
Is it possible to have on_enter method for Toggle component?


VedantParanjape:
Hey, is it possible to add on_enter() for toggle component? Since, this framework doesn't have buttons. Toggle is a pretty good way to emulate buttons. 

I am more than happy to contribute for this feature.

VedantParanjape:
Hey @ArthurSonzogni any updates on this?

ArthurSonzogni:
Hi, thanks for opening this!
We can add on_enter() to match RadioBox and CheckBox.
Do you prefer to do the change? Or I can do it.

Do you think we need to add a "Button" component? I can do it too.

VedantParanjape:
> Hi, thanks for opening this!
> We can add on_enter() in the toggle component. This will match RadioBox and CheckBox.
> Do you prefer to do the change? Or I can do it.

I will make the change, and can we have a identical API for all component, like in RadioBox on_change() actually checks if user pressed enter, and not if the selected changed.

> Do you think we need to add a "Button" component? I can do it too.

With, Toggle component having a on_enter method most probably no.

VedantParanjape:
Hey @ArthurSonzogni, Can we have a identical API for on_change, on_enter for all the components, if yes. I'd like to do that.

Like radio box's on_change is actually on_enter, and some component lack the on_change thing.

Regards,
Vedant

ArthurSonzogni:
We currently have:


| Component | on_change() |on_enter() |
|---|---|---|
| Toggle | X| |
| Menu | X | X |
| RadioBox | X | |
| Input | X | X |
| CheckBox | X |   |
------------------------------------------------------

- on_change() means the data's have changed. The selected data can change by navigating (menu, Toggle), pressing a button like space or enter (RadioBox, CheckBox) or by directly typing something (Input)

- on_enter() was added for Menu and Input. This means "The user has validated the selected data".

I think I left on_enter() in Toggle, because this was copy-pasted from Menu.

I don't think we can make enter to trigger on_enter() instead of on_change() for RadioBox and CheckBox, this would prevent us from using the current behavior.

VedantParanjape:
> I don't think we can make enter to trigger on_enter() instead of on_change() for RadioBox and CheckBox, this would prevent us from using the current behavior.

Current behaviour as in ? sorry, i didn't understand your point.

ArthurSonzogni:
[...]

PR #38, opened 2020-08-04
Added on_enter method to toggle component, and test cases for the same

VedantParanjape:
Solves #37


ArthurSonzogni:
Thanks! This is perfect. Thanks for providing the test.

Issue #45, opened 2020-08-30
Support for RGB Colors


KerDelos:
Hi !

From what I understand there's only 16 colors available in FTXUI for now. Some terminals are able to display a lot more than 16 colors though and it would be great if FTXUI could also support.

I imagine you already now much more than I do about terminal color handling but I'm putting those two link as a reference purpose.
http://jafrog.com/2013/11/23/colors-in-terminal.html
https://tforgione.fr/posts/ansi-escape-codes/
Since I didn't knew how to handle color on a terminal before reading them I figure they can also be useful to somebody else looking at this issue.

**Suggested implementation :**
Transforming the Color enum into a class.
Creating an enum for the color mode. So the Color instance can know if it is a basic color, a 256 color or an RGB color and print the corresponding escape code. (Only supporting RGB colors would be more elegant but I'm not certain it would work the same on every terminal)
Then, depending on how much you want the library to support older code, I can think of two ways to ensure the least amount of external code would have to be rewritten. Either having constructors for the 16 colors already available or having a static instance for each of these colors in the class.

I'd love to work on a pull request for this feature if you're okay with that !

Have a nice day !
Damien.

ps : I've began using FTXUI for a personal project and I like it a lot ! Thank you for creating it :)

ArthurSonzogni:
Thanks for this suggestion!

I would be happy to receive pull request for this feature.
Before starting, do we have a way to check if the current terminal support 16colors/256colors/16Mcolors?
I would suggest reading: https://gist.github.com/XVilka/8346728.
Maybe proceeding like this:
- Checking the env variable $COLORTERM (truecolor => 256
- [fallback] Checking the environment variable $TERM if it contains "256.
- [fallback] Use 16colors.

Alternatively, maybe we can try to read the terminfo?
https://invisible-island.net/ncurses/man/terminfo.5.html

Ideally:
- [ ] On terminal not configured for 256colors, it should fallback gracefully by selecting the closest 16colors.
- [ ] The new API should preserve compatibility with the current FTXUI version.
- [ ] Be cross-platform?

KerDelos:
Cool I'll start working on it then ! Not sure how long it will take me though. It will depend on how much free time I have in the upcoming days/weeks.

Thanks for the links ! I'll take a look and test how we can check color support on terminals.

About the last three points : 
[...]

PR #46, opened 2020-09-02
Rgb color support

KerDelos:
As discussed here #45 this is a pull request to support true colors in FTXUI.

I've managed to change the color enum into a class and make it support three color modes (Basic, Palette256 and TrueColors). Everything worked fine on the terminals I've tested. I've updated the `style_color.cpp` example to display the new color possibilities.

Things that may not be obvious from the github pull request page : I've moved `terminal.hpp` from `src/` to `include/`. I've created a `color.cpp` in `src/screen`. So now there's two `color.cpp` files in the repo but the other is in `src/dom` and cmake seemed okay with that.

About fallbacks and color approximations: 
*The `COLORTERM` environment variable worked fine on my terminal. True color support can thus be tested by `Terminal::CanSupportTrueColors()`. However, my understanding is that `TERM` just contains the name of the terminal and there's no way to be sure how many colors it support without looking at the Terminfo. So since I wasn't sure looking for 256 in `TERM` was enough and i wasn't able to find how to use the terminfo without ncurses there's currently no way to check for color support except for `Terminal::CanSupportTrueColors()`. 
*From what I've seen running tests on a terminal that supports only 8 colors it seemed to approximate and fallback on it's own on its 8 available colors so It didn't seemed worth it to implement an approximation in FTXUI, especially given I wasn't able to properly detect the number of color supported.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+build/
+.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,0 @@
-build/
-.vscode/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,7 @@ find_package(Threads)
 
 add_library(screen
   src/ftxui/screen/box.cpp
+  src/ftxui/screen/color.cpp
   src/ftxui/screen/screen.cpp
   src/ftxui/screen/string.cpp
   src/ftxui/screen/terminal.cpp
@@ -163,7 +164,7 @@ export(TARGETS screen dom component NAMESPACE ftxui::
   FILE ${PROJECT_BINARY_DIR}/ftxui-targets.cmake)
 
 
-if (FTXUI_BUILD_TESTS AND ${CMAKE_VERSION} VERSION_GREATER "3.11.4") 
+if (FTXUI_BUILD_TESTS AND ${CMAKE_VERSION} VERSION_GREATER "3.11.4")
   set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
 
   set(FETCHCONTENT_UPDATES_DISCONNECTED TRUE)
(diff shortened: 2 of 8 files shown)


ArthurSonzogni:
Woohoo Thanks you!
I will take a look at your patches.

ArthurSonzogni:
Hi Damien,

I took a look, your patches were really excellents!

I applied some updates in:
https://github.com/ArthurSonzogni/FTXUI/commit/bcf6d8175b2cd09e0ad38d554f35a3516301340d

For now, I did:
1) Squash the patches.
2) Turn the "static" Color into enums. Add implicit constructors from enum(s) to Color.
3) Reduce the Color Size class size.
4) Give a name to every Palette256 colors;
5) Add Color::HSV function.
6) Add a few more examples.

I still would like to implement "fallback" of colors. So that colors written for truecolor terminal could be displayed in 256terminal and 256terminal into basic terminal.

I can probably finish it this weekend.

KerDelos:
Hi Arthur,

Thanks :) I'm glad I could contribute a little bit!

All your updates looks really nice!  Not sure why I didn't write implicit constructors from enum (like you suggested in the issue) in the first place instead of static objects but you're right it's more elegant this way. 

I would also suggest adding a default constructor to the class because without it, operator[] of std::map doesn't work. Not sure if the default should be a color (like black or white) or some invalid data though. And since there are workarounds to the operator[] it's not that important but I thought I'd mention it.

PS: I realised with the Travis check that I broke the Windows build. Sorry about that !

ArthurSonzogni:
I merged your patch into master.
https://github.com/ArthurSonzogni/FTXUI/commit/dc8c090753ea8f1a9b0d0f2d86fb485b231f7d07

> I would also suggest adding a default constructor to the class because without it, operator[] of std::map doesn't work. Not sure if the default should be a color (like black or white) or some invalid data though. And since there are workarounds to the operator[] it's not that important but I thought I'd mention it.

Done thanks!

> PS: I realised with the Travis check that I broke the Windows build. Sorry about that !

Done thanks!

KerDelos:
thanks !

PR #51, opened 2020-10-15
Fix crash on unset environment variable.

stephanroslen:
My environment doesn't have the variable COLORTERM set. Assigning nullptr to a std::string causes the following exception:
```
a terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
```


ArthurSonzogni:
Thanks for the fix!

This bug was introduced recently in:
https://github.com/ArthurSonzogni/FTXUI/commit/dc8c090753ea8f1a9b0d0f2d86fb485b231f7d07s 

This patch looks perfect.

ArthurSonzogni (approved):

PR #52, opened 2020-10-16
add dependency to Threads into generated cmake config

stephanroslen:
When using FTXUI in my own program with CMake after `find_package(ftxui)` I had to `find_package(Threads)` which indicated, that the cmake config scripts of ftxui didn't model that dependency. I modified the generation of the cmake scripts to yield that dependency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,16 +150,24 @@ if(FTXUI_ENABLE_INSTALL)
   install(DIRECTORY include/ftxui DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 
   include(CMakePackageConfigHelpers)
+  configure_package_config_file(ftxui-config.cmake.in
+    ${CMAKE_CURRENT_BINARY_DIR}/ftxui-config.cmake
+    INSTALL_DESTINATION ${CMAKE_INSTALL_LIBDIR}/ftxui/cmake
+    PATH_VARS CMAKE_INSTALL_INCLUDEDIR
+  )
   write_basic_package_version_file(
-    ftxui-config.cmake
+    ftxui-config-version.cmake
     VERSION ${PACKAGE_VERSION}
     COMPATIBILITY AnyNewerVersion
   )
 
+  install(FILES ${CMAKE_CURRENT_BINARY_DIR}/ftxui-config.cmake
+          DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/ftxui
+  )
   install(EXPORT ftxui-export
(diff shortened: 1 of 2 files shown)


ArthurSonzogni:
Thanks!

I need some time to understand this patch. I will take a look.

Could you show me the error you are getting?

Most dependencies of FTXUI aren't seen any problems. I guess  are using CMake fetchContent:
Like: https://github.com/ArthurSonzogni/ftxui-starter

stephanroslen:
I installed FTXUI and use it with `find_package(ftxui)`. Now I use the following CMakeLists.txt:
```
cmake_minimum_required(VERSION 3.18)
project(ftxui_test)

set(CMAKE_CXX_STANDARD 20)

find_package(ftxui REQUIRED)

add_executable(ftxui_test main.cpp)
target_link_libraries(ftxui_test ftxui::screen ftxui::dom ftxui::component)
```

This yields:
```
CMake Error at CMakeLists.txt:8 (add_executable):
  Target "ftxui_test" links to target "Threads::Threads" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
```

I could workaround this by inserting `find_package(Threads)`into my CMakeLists.txt, but by `find_package(ftxui)` should already find that as a dependency. That's what my modification does - it makes sure, that when using `find_package(ftxui)` it already finds `Threads` with `find_dependency(Threads)`

CMakeLists.txt
```diff
@@ -150,16 +150,24 @@ if(FTXUI_ENABLE_INSTALL)
...
+    ftxui-config-version.cmake
     VERSION ${PACKAGE_VERSION}
     COMPATIBILITY AnyNewerVersion
   )
 
+  install(FILES ${CMAKE_CURRENT_BINARY_DIR}/ftxui-config.cmake
+          DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/ftxui
+  )
   install(EXPORT ftxui-export
-    FILE ftxui-config.cmake
```
ArthurSonzogni, line 160: Could you turn every tabs into 2 spaces?

ArthurSonzogni (commented):
Thanks!

This looks good % fixing tabs -> spaces.

CMakeLists.txt
stephanroslen: oops, sorry, of course.

ArthurSonzogni: Thanks!

ArthurSonzogni (approved):

ArthurSonzogni:
Thank you!

PR #53, opened 2020-10-16
fix documentation example of Screen class

stephanroslen:
`PixelAt(10,10)` was outside of the Screen and there is no c'tor for Dimensions but a static `Create` method.


ArthurSonzogni:
Thanks! LGTM.

ArthurSonzogni (approved):

PR #54, opened 2020-10-16
reset pixel style for newline

stephanroslen:
I observed the problem, that stepping into the next line when printing a screen fills the terminal column with the previous background color:
Bildschirmfoto 2020-10-16 um 18 51 03

My fix yields this:
Bildschirmfoto 2020-10-16 um 18 58 04

This is my example code: 
```cpp
#include <iostream>

int main(void) {
  using namespace ftxui;
  Element document = graph([](int x, int y){
    std::vector<int> result(x, 0);
    for (int i{0}; i < x; ++i) {
      result[i] = ((3 * i) / 2) % y;
    }
    return result;
  }) | color(Color::Red) | border | color(Color::Green) | bgcolor(Color::DarkBlue);

  auto screen = Screen::Create(
      Dimension::Fixed(80),
      Dimension::Fixed(10)
  );
  Render(screen, document);
  std::cout << screen.ToString() << '\n';
}
```

Btw. this library is really great! I love its API and power. I've been looking for a lib like this for some time. :)

### DIFF
--- a/src/ftxui/screen/screen.cpp
+++ b/src/ftxui/screen/screen.cpp
@@ -153,10 +153,13 @@ std::string Screen::ToString() {
   std::wstringstream ss;
 
   Pixel previous_pixel;
+  Pixel final_pixel;
 
   for (int y = 0; y < dimy_; ++y) {
-    if (y != 0)
+    if (y != 0) {
+      UpdatePixelStyle(ss, previous_pixel, final_pixel);
       ss << '\n';
+    }
     for (int x = 0; x < dimx_;) {
       auto& pixel = pixels_[y][x];
       wchar_t c = pixel.character;
@@ -166,7 +169,6 @@ std::string Screen::ToString() {
     }
   }
 
-  Pixel final_pixel;
   UpdatePixelStyle(ss, previous_pixel, final_pixel);
 
   return to_string(ss.str());


stephanroslen:
Mhh, that failed CodeFactor check is not referring to a function I changed...

ArthurSonzogni:
Thanks! LGTM.

I can't reproduce the issue on my terminal. I guess it depends on what terminal you use.

> Mhh, that failed CodeFactor check is not referring to a function I changed...

Yes, I probably should remove this bot.

ArthurSonzogni (approved):

ArthurSonzogni:
> Btw. this library is really great! I love its API and power. I've been looking for a lib like this for some time. :)

Thanks you!

I am quite happy about screen/ and dom/ directories.
I am still unsure about the API for in component/ (interactive UI controllable from the keyboard)

stephanroslen:
> I am still unsure about the API for in component/ (interactive UI controllable from the keyboard)

I just started to understand how it really works - I find the idea brilliant, to let the interactive components generate dom elements. It took me some time to grasp it.

ArthurSonzogni:
>> I am still unsure about the API for in component/ (interactive UI controllable from the keyboard)

>I just started to understand how it really works - I find the idea brilliant, to let the interactive components generate dom elements. It took me some time to grasp it.

Yes, compared the dom/ module, the component/ module is not easy to start with. I haven't really found a way to make it immediate to understand. Maybe I should write a tutorial.
At least, reading all the examples should help a bit.

PR #57, opened 2020-10-24
Make input cursor position public state

scaryrawr:
- Cursor position is very well protected in the render function
- Making it public enables scenarios like
  - Tab completion
  - Inheritance for things like suggestions/previews with tab completion


ArthurSonzogni (approved):
LGTM! Thanks!

Issue #55, opened 2020-10-23
Esc-key doesn't trigger on single press?


scaryrawr:
I'm loving FTXUI, but I'm noticing that when I try to do something like:

```c++
bool CustomComponent::OnEvent(Event event)
{
    if (Event::Escape == event)
    {
        // Never hit.
        on_exit();
        return true;
    }
}
```

It looks like there's a `ParseESC` that waits for other possible combinations with the escape key.

I'm also wondering if I just have it set up wrong...

My main loop (I'm making a launcher...):
```c++
	auto screen = ftxui::ScreenInteractive::TerminalOutput();

	tofi::Modes modes;
	modes.emplace_back(std::make_unique<tofi::modes::drun>());
	modes.emplace_back(std::make_unique<tofi::modes::run>());

	tofi::Tofi tofi{std::move(modes)};
	int exit{};
	tofi.on_exit = [&exit, &screen](int code) {
		exit = code;
		screen.ExitLoopClosure()();
	};

	screen.Loop(&tofi);
```

I see if I hit Esc 3 times I'll get an input of "\033\033\033", but I can't seem to get "\033".

Thanks!

ArthurSonzogni:
Thanks for reporting this!

From the ./build/example/util/print_key_press, you can see what sequence is emitted for every keys.

It sounds like the sequence of bytes sent for one key is a prefix of others. This make it impossible to correctly decode the sequence.
I mostly used this resource:
https://invisible-island.net/xterm/ctlseqs/ctlseqs.html

For instance:
- ESC => ( 27 )
- F1 => ( 27 79 80 )

As a result, the current implementation, after reading the first (27), tries to consume more from the input.

We need to find how application are handling it. It sounds they are relying on some kind of timeout (search for escape-time)

I am not quite sure how to handle this.

ArthurSonzogni:
I have several ideas. I will try fixing this.

ArthurSonzogni:
And this is fixed !

PR #58, opened 2020-10-24
Fix parsing of keys that are prefix of others.

ArthurSonzogni:
The ESC key generates sequences that are prefix of others. For instance:
- ESC => [27]
- F1 =>  [27, 79, 8]

As a result, we can't generate the ESC event when receiving [27],
because it might be the start of the [27, 79, 8] sequence (or not).

Application usually applies a timeout to help detecting the ESC key.
This patch introduce a timeout. It is set to 50ms.

Bug: https://github.com/ArthurSonzogni/FTXUI/issues/55


ArthurSonzogni:
+ @scaryrawr FYI

This should fix the issue with the ESC key.

I need to confirm this compile on Windows before landing this.

scaryrawr:
Seems to be working well!

Issue #60, opened 2020-11-14
Build fails if this library is used in a non-git folder


VedantParanjape:
So, If this library is used in a folder which is not a git repository build will fail

CMakeLists.txt

```cmake
# Specify minimum cmake version required
cmake_minimum_required(VERSION 3.4)

# Download ftxui library
include(FetchContent)

set(FETCHCONTENT_UPDATES_DISCONNECTED TRUE)
FetchContent_Declare(ftxui
  GIT_REPOSITORY https://github.com/ArthurSonzogni/ftxui
)

FetchContent_GetProperties(ftxui)
if(NOT ftxui_POPULATED)
  FetchContent_Populate(ftxui)
  add_subdirectory(${ftxui_SOURCE_DIR} ${ftxui_BINARY_DIR} EXCLUDE_FROM_ALL)
endif()

# Find Threads Library
find_package( Threads )

# CMake project description and project specific settings
project(simppru-console
        VERSION 1.0
        DESCRIPTION "Console for simpPRU"
        HOMEPAGE_URL "https://github.com/VedantParanjape/simpPRU/")
set_property(GLOBAL PROPERTY USE_FOLDERS ON)
set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_SOURCE_DIR}/bin")

# Set pru-gcc as the compiler

# Specify Build type
if(NOT CMAKE_BUILD_TYPE)
  set(CMAKE_BUILD_TYPE Release)
endif()
  
######## SOURCES ########
# Add source files and header files
file(GLOB PROJECT_HEADERS *.hpp)
file(GLOB PROJECT_SOURCES *.cpp)
set(PROJECT_FILES
    ${PROJECT_HEADERS}
    ${PROJECT_SOURCES})

######## TARGETS ########
# Add executables
add_executable(${PROJECT_NAME}-${CMAKE_PROJECT_VERSION} ${PROJECT_FILES})
set_target_properties(${PROJECT_NAME}-${CMAKE_PROJECT_VERSION} PROPERTIES CXX_STANDARD 17)
target_include_directories(${PROJECT_NAME}-${CMAKE_PROJECT_VERSION} PUBLIC 
  $<BUILD_INTERFACE:${simppru_SOURCE_DIR}/include/console>
  $<BUILD_INTERFACE:${ftxui_SOURCE_DIR}/include>
  $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>)

[...]

PR #61, opened 2020-11-14
Fixed issue (#60) - Compilation failed in non-git projects

VedantParanjape:
Compilation failed if ftxui was used in projects that were not git repositories. Since, in CMakeLists.txt working directory to execute command was set as the parent directory and not the FTXUI project directory.

Changed the working directory variable to ${CMAKE_CURRENT_SOURCE_DIR}, so as to get the version of FTXUI and not that of parent project

Fixes issue #60


ArthurSonzogni (approved):
Looking good!

ArthurSonzogni:
Thanks you!

PR #65, opened 2020-12-23
Fix terminal color detection support.

ArthurSonzogni:
There was many obvious bugs, discovered here:
https://github.com/Gusabary/SAIL/blob/ad2a085469d80cfdcccd13abc5efcbabf1154cd4/Open-Source-Project-Anatomy/FTXUI.md

### DIFF
--- a/src/ftxui/screen/terminal.cpp
+++ b/src/ftxui/screen/terminal.cpp
@@ -41,15 +41,19 @@ const char* Safe(const char* c) {
   return c ? c : "";
 }
 
+bool Contains(const std::string& s, const char* key) {
+  return s.find(key) != std::string::npos;
+}
+
 static bool cached = false;
 Terminal::Color cached_supported_color;
 Terminal::Color ComputeColorSupport() {
   std::string COLORTERM = Safe(std::getenv("COLORTERM"));
-  if (COLORTERM.compare("24bit") || COLORTERM.compare("trueColor"))
+  if (Contains(COLORTERM, "24bit") || Contains(COLORTERM, "truecolor"))
     return Terminal::Color::TrueColor;
 
   std::string TERM = Safe(std::getenv("TERM"));
-  if (COLORTERM.compare("256") || COLORTERM.compare("256"))
+  if (Contains(COLORTERM, "256") || Contains(TERM, "256"))
     return Terminal::Color::Palette256;
 
   return Terminal::Color::Palette16;


ArthurSonzogni:
+ @Gusabary FYI. Thanks!

src/ftxui/screen/terminal.cpp
```diff
@@ -41,15 +41,19 @@ const char* Safe(const char* c) {
...
 Terminal::Color cached_supported_color;
 Terminal::Color ComputeColorSupport() {
   std::string COLORTERM = Safe(std::getenv("COLORTERM"));
-  if (COLORTERM.compare("24bit") || COLORTERM.compare("trueColor"))
+  if (Contains(COLORTERM, "24bit") || Contains(COLORTERM, "truecolor"))
     return Terminal::Color::TrueColor;
 
   std::string TERM = Safe(std::getenv("TERM"));
-  if (COLORTERM.compare("256") || COLORTERM.compare("256"))
+  if (Contains(COLORTERM, "256") || Contains(TERM, "256"))
```
daminetreg, line 56: I just saw the notification on github, isn't in principle checked through the `isatty` call on stdout ? (works on windows too)

daminetreg: Just based on what what I saw in some termcolor lib I use https://github.com/ikalnytskyi/termcolor/blob/ae1fa0d77790d68a441a09bfacd5654cf2c9c6e5/include/termcolor/termcolor.hpp#L539

But it seems to be an assumption more that a true info. Sorry for disturbing here ^^

ArthurSonzogni: Thank!
I think there are two orthogonal things to check:
- Whether the output is plugged toward a terminal (isatty)
- What kind of colors are supported by the terminal (2^1, 2^4, 2^8, 2^24)

With FTXUI, for now, I expect both input and output to be plugged toward a terminal. I don't really support other modes. Maybe I should in the future?

ArthurSonzogni:
FYI. I am trying to fallback colors palette depending on what the terminal support.
Here is what I got::
[asciicast](https://asciinema.org/a/380833)

src/ftxui/screen/terminal.cpp
daminetreg: Yes makes sense 👍 I think it's safe to assume it's a terminal.

daminetreg:
> FYI. I am trying to fallback colors palette depending on what the terminal support.
> Here is what I got::
> [asciicast](https://asciinema.org/a/380833)

Beautiful 😍

Gusabary:
> * @Gusabary FYI. Thanks!

my pleasure 🤣

Gusabary (commented):
cool fix

PR #66, opened 2020-12-23
Color fallback

ArthurSonzogni:
On terminal not supporting some color palette, fallback toward the closest one.


ArthurSonzogni:
Some pictures:
[asciicast](https://asciinema.org/a/380837)
[asciicast](https://asciinema.org/a/380833)

+@daminetreg FYI

PR #69, opened 2021-02-01
Fixed infinite loop in Screen::ToString() for non-printable chars

gfannes:
I decided to replace the non-printable character with a `space` and continue over a hang or throwing an exception: a CPU-burning hang is not nice and the rest of the code seems to not use exceptions.


ArthurSonzogni (approved):
Thanks! This is looking good!

PR #71, opened 2021-02-10
Build with -Wmissing-declarations

robinlinden:
This flag is used to find global functions defined without a previous
declaration. Usually it spots accidental globals, but in this case it
was just missing headers.

```
Element separator()
Element separator(Pixel)
Element text(std::wstring)
Element vtext(std::wstring)
```
are all declared in `ftxui/dom/elements.hpp` which was missing from
the includes in the .cpp-files where they were defined.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -142,6 +142,7 @@ foreach(lib screen dom component)
     target_compile_options(${lib} PRIVATE "-Wextra")
     target_compile_options(${lib} PRIVATE "-pedantic")
     target_compile_options(${lib} PRIVATE "-Werror")
+    target_compile_options(${lib} PRIVATE "-Wmissing-declarations")
     target_compile_options(${lib} PRIVATE "-Wno-sign-compare")
   endif()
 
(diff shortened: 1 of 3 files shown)


ArthurSonzogni (approved):
This is looking good! Thanks!

ArthurSonzogni:
Thanks!
Out of curiosity, what lead you to need this warning in particular?

robinlinden:
Nothing in particular. I just have it set globally in general to catch accidental global functions, and this saves me having to exclude FTXUI when setting it. 😁

I am looking into a real bug that I'm 97% sure is in FTXUI though. It seems to drop a word in the middle of a sentence created like `hflow(paragraph(mystr))` which is a bit odd. The dropped word is right where the sentence wraps in the terminal, but I don't have a minimal repro or anything, so I figured I'd look into it more before opening an issue (or PR fixing it if I work it out myself.)

It reliably reproduces every run of my project, but the tui is generated at runtime and I couldn't find a way to dump the node tree so it's some manual work I haven't done yet to make a minimal repro.

PR #74, opened 2021-03-12
[Win] Fixed crash for intensive sending of keyboard events

jdfa:


ArthurSonzogni:
Hi,
Thanks for this fix!

Could you provide a description of your patch? I don't really understand what was the bug.
If you look at this example:
https://docs.microsoft.com/en-us/windows/console/reading-input-buffer-events
You see the size is in number of elements, not in bytes. So I am wondering if you patch is correct.

jdfa:
Hi,

Sorry, I had to explain this at first place.

I was facing a lot of crashes, even on example applications. They appeared when I was pressing keyboard arrows pretty fast (5-6 times per second). It was looking like this:
image

Issue was that vector with records has capacity of 1 but we are allowing to write 20 items. In my scenario it lead to heap corruption.

That issue was related to wrong usage of ReadConsoleInput, Here is some info from [documentation](https://docs.microsoft.com/en-gb/windows/console/readconsoleinput#parameters):

_nLength [in]
The size of the array pointed to by the lpBuffer parameter, **in array elements.**_

So it's should be simply a size of array.

Even in your example it is the same, there is no need to calculate size in bytes.

ArthurSonzogni:
Ooops, sorry. I thought your patch was doing the opposite.
Thanks!

ArthurSonzogni (approved):

ArthurSonzogni:
Thanks you!

If you find other bugs, feel free to submit PR or fill issues. They are greatly appreciated !
I don't own a Windows OS, so it is pretty difficult for me.

PR #75, opened 2021-03-16
Swapped order of hidding cursor and setting alternative screen

jdfa:
Text cursor is visible when I use `ScreenInteractive::Fullscreen()`. Not sure is it reproducible on Linux but that's the case on Windows. I was using menu example and changed only screen mode:

drawing

I assume that usage of alternative screen command resets cursor hiding. Swap of those commands fixed that issue:

drawing

Need to check alternative screen functionality, looks like it was affected


ArthurSonzogni (approved):
I cannot reproduce your problem on my terminal. I am using kitty on Linux.
I don't own a windows OS, so your patch is greatly appreciated. Thanks you!

PR #80, opened 2021-03-22
Fix MSVC Warning D9025

YuzukiTsuru:
`cl : Command line warning D9025 : overriding '/W3' with '/W4'`


ArthurSonzogni (approved):

ArthurSonzogni:
Thanks!
Yes, I wish I could use the most aggressive warning, but I don't own a windows OS, so it is quite difficult for me to fix every warnings. I will at some point, but in the meantime, I guess your patch is a good compromise to avoid spamming windows users.

LGTM!

PR #111, opened 2021-06-12
Compatify the Screen memory.

ArthurSonzogni:
Instead of storing all the booleans with 5 bytes, use a bit field.
The size of a Pixel has been reduced by 25%, from 20 byte to 15 bytes.

### DIFF
--- a/include/ftxui/screen/screen.hpp
+++ b/include/ftxui/screen/screen.hpp
@@ -18,14 +18,21 @@ using Element = std::shared_ptr<Node>;
 /// @brief A unicode character and its associated style.
 /// @ingroup screen
 struct Pixel {
-  wchar_t character = U' ';
-  bool blink = false;
-  bool bold = false;
-  bool dim = false;
-  bool inverted = false;
-  bool underlined = false;
   Color background_color = Color::Default;
   Color foreground_color = Color::Default;
+  wchar_t character = U' ';
+  bool blink : 1;
+  bool bold : 1;
+  bool dim : 1;
+  bool inverted : 1;
+  bool underlined : 1;
+
+  Pixel()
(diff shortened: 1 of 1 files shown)


include/ftxui/screen/screen.hpp
```diff
@@ -18,14 +18,21 @@ using Element = std::shared_ptr<Node>;
...
   Color background_color = Color::Default;
   Color foreground_color = Color::Default;
+  wchar_t character = U' ';
+  bool blink : 1;
+  bool bold : 1;
+  bool dim : 1;
+  bool inverted : 1;
+  bool underlined : 1;
+
+  Pixel()
```
mark2185, line 30: The values can be defined in the declaration, there is no need for a constructor.

E.g. `bool blink : 1 = false;` - [godbolt link](https://godbolt.org/z/n6Wx1YzGo)

ArthurSonzogni: Thanks!

This was done on purpose, because FTXUI support C++17 and this will be available on C++20:
https://godbolt.org/z/e4ajhcYvT
```
warning: default member initializers for bit-fields only available with '-std=c++20' or '-std=gnu++20' [-Wc++20-extensions]
```

(I had this problem 3 days ago)
https://chromium-review.googlesource.com/c/chromium/src/+/2947501

mark2185: Ah, apologies for the noise then, I just got the mail that a PR was opened so I wanted to help, I didn't get the chance to study the repo in depth yet.

ArthurSonzogni: ;-)
It is nice to see there are some people watching. Thank you!

PR #116, opened 2021-06-16
Set git_version when .git folder is missing

myd7349:
If we download the FTXUI source archive from the GitHub website, then the `.git` folder will be missing.
In such case, CMake configuration will fail:
> git found
> fatal: not a git repository (or any of the parent directories): .git
> CMake Error at CMakeLists.txt:21 (project):
>   VERSION "0.6." format invalid.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.0)
 
 find_package(Git QUIET)
-if (Git_FOUND)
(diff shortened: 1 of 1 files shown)


ArthurSonzogni (approved):
LGTM! Thanks!

myd7349:
Please do not merge yet. Let's if this patch can fix Linux and macOS build errors with `vcpkg`:
https://github.com/microsoft/vcpkg/pull/18360

myd7349:
Hi. @ArthurSonzogni I think this PR is ready now.

Issue #126, opened 2021-06-25
Feature request: A menu of components?


ArthurSonzogni:
It would be nice adding a Menu of components, instead of being a menu of focusable text.

This would be useful for beagle-config, which was using a group of 3 submenu.

SAtacker:
Hi @ArthurSonzogni  how about 
[![asciicast](https://asciinema.org/a/YLWllRq0PIx5s1cjzetr6VuW4.svg)](https://asciinema.org/a/YLWllRq0PIx5s1cjzetr6VuW4)

SAtacker:
If that's okay, I can send a PR right now.

PR #131, opened 2021-06-26
Menu of components

SAtacker:
Issue #126 
[asciicast](https://asciinema.org/a/YLWllRq0PIx5s1cjzetr6VuW4)


ArthurSonzogni:
Thank you for trying to implement this!

I am going to push a few changes over yours that I would like to have.

ArthurSonzogni:
Hi @SAtacker 

Could you please take a look?

Instead of introducing a new class and template, I chose to be able to provide Container::Vertical and Container::Horizontal able to get an external |selector| like Container::Tab.

This should fit our purpose I believe.

SAtacker:
Thank you @ArthurSonzogni 
That's nicer than menu of components and also it made more sense to modify containers and not `menu` component.
I haven't looked it locally, will get back here.

Issue #136, opened 2021-07-03
[BUG] Input fields get random symbols like "[B[AA1" or "B" and others pasted in on every action


Creapermann:
Input fields get random symbols like "[B[AA1" or "B" and others pasted in on every action like typing, using the arrow keys and moving the mouse

ArthurSonzogni:
Hello, what kind of terminal emulator are you using?

ArthurSonzogni:
In particular, what sequence of byte it sent when you are using the example

./example/util/print_key_press
and use your keyboard?

You should get those numbers:
```
│( 27 91 65 ) -> (special)                                                                               │
│( 27 91 68 ) -> (special)                                                                               │
│( 27 91 66 ) -> (special)                                                                               │
│( 27 91 67 ) -> (special)                                                                               │
```

You can try it online against the xterm.js terminal emulator and check you receive the same sequences:
https://arthursonzogni.com/FTXUI/examples/?file=./util/print_key_press.js

Creapermann:
It changes how extreme it is, sometimes even more symbols than here get pasted in. https://streamable.com/mg7g85
All actions make this happen, deleting, switching with arrow keys or just writing.

AhmedMostafa16:
> 
> 
> Hello, what kind of terminal emulator are you using?

Windows CMD. I have tried PowerShell but the same.

Creapermann:
> 
> 
> > Hello, what kind of terminal emulator are you using?
> 
> Windows CMD. I have tried PowerShell, but the same.

@ArthurSonzogni  we are working on the same project btw, so i m using Windows CMD too. Btw, its not the same symbols your https://arthursonzogni.com/FTXUI/examples/?file=./util/print_key_press.js shows me

Creapermann:
@ArthurSonzogni, do u have any idea?

ArthurSonzogni:
I finally was able to test on Windows.
There are several problems I tried to mitigate with:
https://github.com/ArthurSonzogni/FTXUI/pull/138

However I was not able to reproduce your problems. I just saw you made a video. This help a lot. Thanks!

Creapermann:
> 
> 
> I finally was able to test on Windows.
> There are several problems I tried to mitigate with:
> #138
> 
> However I was not able to reproduce your problems. I just saw you made a video. This help a lot. Thanks!

What should we do now? If you check my github account, you can find my project called "PW-Manager" clone it and run it, you ll see the errors too

[...]

Issue #135, opened 2021-06-29
Document how to enable truecolor on Windows


MichaelGoulding:
The code looks for the environment variable
```
COLORTERM
```
to enable true color.

Windows 10 has supported true color in the console for a while now, but FTXUI doesn't do the right thing without the environment variable set:

![image](https://user-images.githubusercontent.com/18077959/123858633-9b00f980-d8d8-11eb-9cd3-ae242058b3e2.png)

ArthurSonzogni:
I am sorry this terminal on Windows doesn't set the COLORTERM variable.

Maybe there standard ways in the specification:
https://invisible-island.net/xterm/ctlseqs/ctlseqs.html
to query directly the terminal support for colors.

The cost would be some communication with the terminal via stdin/stdout.
Maybe 
```cpp
ftxui::QueryColorSupport()
```

ArthurSonzogni:
Hey @MichaelGoulding,

Do you think we should close this issue, given the latest patch?

PR #138, opened 2021-07-04
Implement Fallback for microsoft's terminals.

ArthurSonzogni:
Implement Fallback for microsoft's terminals.

I finally got access to a computer using the Microsoft's Windows OS.
That's the opportunity to find and mitigate all the problems
encountered. This patch:

1. Introduce an option and a C++ definition to enable fallback for
   Microsoft's terminal emulators. This allows me to see/test the
   Microsoft output from Linux. This also allows Windows users to remove
   the fallback and target non Microsoft terminals on Windows if needed.

2. Microsoft's terminal suffer from a race condition bug when reporting
   the cursor position:
   https://github.com/microsoft/terminal/pull/7583.
   The mitigation is not to ask for the cursor position in fullscreen
   mode where it isn't really needed and request it less often.
   This fixes: https://github.com/ArthurSonzogni/FTXUI/issues/136

3. Microsoft's terminal do not handle properly hidding the cursor. Instead
   the character under the cursor is hidden, which is a big problem. As
   a result, we don't enable setting the cursor to the best position for
   [input method editors](https://en.wikipedia.org/wiki/Input_method),
   It will be displayed at the bottom right corner.
   See:
   - https://github.com/microsoft/terminal/issues/1203
   - https://github.com/microsoft/terminal/issues/3093

4. Microsoft's terminals do not provide a way to query if they support
   colors. As a fallback, assume true colors is supported.
   See issue:
   - https://github.com/microsoft/terminal/issues/1040
   This mitigates:
   - https://github.com/ArthurSonzogni/FTXUI/issues/135

5. The "cmd" on Windows do not properly report its dimension. Powershell
   works correctly. As a fallback, use a 80x80 size instead of 0x0.

6. There are several dom elements and component displayed incorrectly,
   because the font used is missing several unicode glyph. Use
   alternatives or less detailled one as a fallback.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,17 @@ option(FTXUI_BUILD_TESTS "Set to ON to build tests" OFF)
 option(FTXUI_BUILD_TESTS_FUZZER "Set to ON to enable fuzzing" OFF)
 option(FTXUI_ENABLE_INSTALL "Generate the install target" ON)
 
+set(FTXUI_MICROSOFT_TERMINAL_FALLBACK_HELP_TEXT "On windows, assume the \
+terminal used will be one of Microsoft and use a set of reasonnable fallback \
+to counteract its implementations problems.")
+if (WIN32)
+  option(FTXUI_MICROSOFT_TERMINAL_FALLBACK
+    ${FTXUI_MICROSOFT_TERMINAL_FALLBACK_HELP_TEXT} ON)
+else()
+  option(FTXUI_MICROSOFT_TERMINAL_FALLBACK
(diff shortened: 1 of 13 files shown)


ArthurSonzogni:
Hey @MichaelGoulding,
I finally got a way to test FTXUI on Windows. There are a few problems with this terminal emulator that I do not encounter on the other terminal emulators.

Does the last thing about truecolor capability detection looks like a reasonable thing to do on Windows? Could you please take a  look?

ArthurSonzogni:
@Creapermann. I believe this should mitigate your issue.

PR #140, opened 2021-07-04
Rename library for vcpkg.

ArthurSonzogni:
Instead of putting the archive as:
- <lib_dir>/ftxui/libcomponent.a
- <lib_dir>/ftxui/libdom.a
- <lib_dir>/ftxui/libscreen.a

Use:
- <lib_dir>/libftxui-component.a
- <lib_dir>/libftxui-dom.a
- <lib_dir>/libftxui-screen.a

This fixes an issue users have with VCPKG.

See:
https://github.com/microsoft/vcpkg/issues/16327#issuecomment-868942370


ArthurSonzogni:
Hi @myd7349, could you please take a look?

myd7349:
Hi! @ArthurSonzogni Sorry for the long delay.
I tested on my PC this afternoon, and it works as expected.

https://github.com/microsoft/vcpkg/pull/18360 is already merged, without this patch however.
I think we can update vcpkg's port file later.

PR #144, opened 2021-07-07
On change option for radiobox

gokulmaxi:


ArthurSonzogni:
Thanks!

Soon, I would like to deprecate the non functional use of the Component. I would like people to stop using the RadioboxBase and  RadioboxBase::From(Component) for instance.

So I would like to provide everything via the function you modified. However, I don't want to break the API. So I would like to wrap every "Optional" attribute instead and object.
Something like:
```cpp
Component Radiobox(const std::vector<std::wstring>* entries, int* selected_, Ref<RadioboxOption> options)
```

Would you mind waiting for me introducing this before landing your patch?

gokulmaxi:
I was wondering how to implement on_change and found nothing to my surprise.So i started creating this PR and implemented it in one of my project [vantage](https://github.com/gokulmaxi/vantage)

gokulmaxi:
I am happy and open to help with your patches

PR #150, opened 2021-07-13
Component example showing composition of components

cmorganBE:
This approach can allow for UI sections that are more modular and reusable


cmorganBE:
@ArthurSonzogni here is the second of the two PRs that is intended to show how to build modular components up. I use this technique a bunch in my first ftxui app to group interactive controls and elements into 'panes' or whatever you might call them.

ArthurSonzogni:
Nice!
Thank you!

Issue #149, opened 2021-07-13
Compile error on cygwin


zhangtianli2006:
![error img](https://user-images.githubusercontent.com/49156174/125386565-141b4900-e3cf-11eb-8b9e-3e15380ee630.PNG)

On windows7
gcc version 10.2.0

ArthurSonzogni:
Hmm, that's true.
I followed the UTF8 spec a bit too closely.

Do you want to submit a PR removing this?

ArthurSonzogni:
I will make a PR soon tonight.

ArthurSonzogni:
This might be fixed.
@zhangtianli2006, could you please double check?

PR #151, opened 2021-07-14
Resolve compile error on cygwin.

ArthurSonzogni:
This should solve issue:
https://github.com/ArthurSonzogni/FTXUI/issues/149


cmorganBE (dismissed):
This looks like the correct approach. Checked that its ok that the low bits of 'value' are modified as you shift the content to the left when parsing (this is how utf-8 woks). Type of uint32_t makes sense as that's the max size of a utf-8 code point. I'm guessing this wouldn't have shown up as a functional issue in many cases due to the typical use of code points >= 65535.

ArthurSonzogni:
Thanks for taking a look! Yes, that's because on some platform the size of wchar_t may differ.

BTW, there is a fuzzer:
```
FTXUI_BUILD_TESTS_FUZZER
```
to ensure we parse correctly events.

PR #152, opened 2021-07-14
Allows components to remove a child or access to children in general

felixjulianheitmann:
If I create a Container with a list of children to be displayed and want to delete an entry, I would like to have a functionality corresponding to the add function. 
Furthermore, I added a begin/end function to ComponentBase to allow iteration over its children. The Remove function can then be conformant with vector::erase.

### DIFF
--- a/include/ftxui/component/component_base.hpp
+++ b/include/ftxui/component/component_base.hpp
@@ -30,6 +30,7 @@ class ComponentBase {
   ComponentBase* Parent();
   void Add(Component children);
   void Remove(Components::iterator child);
+  void Clear();
   // Renders the component.
   virtual Element Render();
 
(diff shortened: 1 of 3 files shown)


felixjulianheitmann:
I also added a Clear function to remove all children from a Component.

ArthurSonzogni:
Thanks!

At this point, we might want to expose:
```cpp
  Components Children() { return children_; }
```
But this won't really work. There is an issue with your patch as well. The child also keep a reference to its parent. If we remove it without clearing the reference, then this will cause a pointer to be dangling.

I will submit some tests and update the implementation.

felixjulianheitmann:
I get that full access would impose problems due to the `parent<->child` relations. Wouldn't a call to a childs `Detach()` method before erasing the pointer from the children vector suffice?

But I'll gladly wait. Thanks for the useful library

ArthurSonzogni:
Hey @felixjulianheitmann, I updated your patch. Could you please take a look?

felixjulianheitmann:
Hi @ArthurSonzogni, thank you for the fast help. I slightly edited some comments but the code works fine for me.

ArthurSonzogni:
Merged!

Thanks Felix!

PR #156, opened 2021-07-16
Modify CMake usage to enable C++17

tusharpm:
The `target_compile_features` command allows setting PUBLIC compile features for the library, making it implicit for the applications linking it. This reduces the CMake boilerplate required to set up a dependent project (eg. ftxui-starter).


ArthurSonzogni (approved):
Thank you!
This looks good.

PR #157, opened 2021-07-17
Remove explicit default destructors

tusharpm:
From CppCoreGuidelines:
- Rule of Zero: [C.20: If you can avoid defining default operations, do](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-zero).
- [C.52: Use inheriting constructors to import constructors into a derived class that does not need further explicit initialization](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c52-use-inheriting-constructors-to-import-constructors-into-a-derived-class-that-does-not-need-further-explicit-initialization).
- [DRY](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-DRY) forward and using declarations.

Miscellaneous:
- Fix `format.sh` to output examples with normalised paths in sorted order.


ArthurSonzogni (approved):
Looking great! Thanks!

PR #160, opened 2021-07-19
Miscellaneous refactoring

tusharpm:
**ContainerBase:**
- Extract Vertical, Horizontal and Tab containers to subclasses of ContainerBase.
- Preserve specific behaviour using virtual methods.

**NodeDecorator:**
- Collect the usage of `unpack` from NodeDecorator subclasses into single constructor.

**Miscellaneous:**
- Replace Graph functor in homescreen example with lambda.
- Reduce duplication of logic for Box Contains point from `Screen::PixelAt`.

### DIFF
--- a/examples/component/homescreen.cpp
+++ b/examples/component/homescreen.cpp
@@ -24,26 +24,19 @@ int main(int argc, const char* argv[]) {
 
   int shift = 0;
 
-  class Graph {
-   public:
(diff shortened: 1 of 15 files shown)


ArthurSonzogni (approved):
Wow. Thanks!

This is looking perfect to me.

PR #162, opened 2021-07-20
Deduplicate logic in ComponentBase members

tusharpm:
- Invoke `DetachAllChildren` from `~ComponentBase`
  - *Note: This reverses the order of Detach calls for children of the component. The previous order is not documented.*
- Define `Focused` using `Active`
- Compact `TakeFocus` loop code
- Add const for `Parent`, `Active` and `Focused`


ArthurSonzogni (approved):
Thanks! LGTM!

Really appreciated!

PR #165, opened 2021-07-23
Install graphviz to let Doxygen generate graph.

ArthurSonzogni:
+CC @tusharpm


tusharpm:
Nice catch! I had missed this as I didn't find any graphs in the docs.

PR #170, opened 2021-07-30
Refactor examples list in CMake

### LINKED ISSUE #169: SharedArrayBuffer is deprecated outside of cross-origin isolated environments.
Since @tusharpm patch:
https://github.com/ArthurSonzogni/FTXUI/pull/166#pullrequestreview-714337259
The documentation and webassembly examples are automatically deployed on https://arthursonzogni.github.io instead of https://arthursonzogni.com

The examples are using webassembly threads and shared memory. This requires in Firefox and Chrome >=92 to be in cross-origin isolated environments using those two headers:
```
cross-origin-embedder-policy: require-corp
cross-origin-opener-policy: same-origin
```

But I don't own arthursonzogni.github.io and can't configure the headers.

In the meantime we should revert back to https://arthursonzogni.com/FTXUI/examples/ so that Firefox and Chrome-beta users can use them.

tusharpm:
- Remove duplicate list of examples from `index.html`.
  - Add CMake logic to collect the list in a global property.
  - Move print_key_press example from util to component for convergence.
- Replace shell script to serve examples with python script.
  - Add COEP and COOP headers in the HTTP server.
- Revert examples link from GitHub pages to arthursonzogni.com.
  - Closes #169.


ArthurSonzogni (approved):
Thanks! LGTM!

PR #171, opened 2021-08-01
Extract common struct Dimensions from Terminal

tusharpm:
- Convert `Dimension` to namespace to allow defining `Fit` method from dom.
  - Use `Dimensions` extracted from `Terminal` as replacement `struct`.
- Convert `Terminal` to namespace as it only defines static members.
- Remove dom references from screen library (circular dependency).


ArthurSonzogni (approved):
LGTM! Thanks!

PR #175, opened 2021-08-06
Rounded border

ArthurSonzogni:


src/ftxui/component/renderer.cpp
```diff
@@ -69,6 +62,48 @@ Component Renderer(Component child, std::function<Element()> render) {
...
+///
+/// ### Example
+///
+/// ```cpp
+/// auto screen = ScreenInteractive::TerminalOutput();
+/// auto renderer = Renderer([] (bool focused) {
+///   if (focused)
+///     return text("My interface") | inverted;
+///   else
+///     return text("My interface") | inverted;
```
tusharpm, line 78: This comment doesn't clarify the usage of the focused condition.
Consider differentiating between the two branches with a different decorator.

src/ftxui/component/renderer.cpp
```diff
@@ -69,6 +62,48 @@ Component Renderer(Component child, std::function<Element()> render) {
...
+      if (event.is_mouse() && box_.Contain(event.mouse().x, event.mouse().y)) {
+        if (!CaptureMouse(event))
+          return false;
+
+        TakeFocus();
+      }
+
+      return false;
+    }
+    Box box_;
```
tusharpm, line 100: The default constructor for Box doesn't initialize the members.
Consider initializing `box_` explicitly to avoid reads before initialization.

Ref: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-always

src/ftxui/dom/border.cpp
```diff
@@ -12,7 +12,8 @@
 
 namespace ftxui {
 
-static wchar_t simple_border_charset[] = L"┌┐└┘─│┬┴┤├";
+//static wchar_t simple_border_charset[] = L"┌┐└┘─│┬┴┤├";
```
tusharpm, line 15: nit: unused comment.

ArthurSonzogni: Indeed. I turned it into a TODO.

src/ftxui/component/renderer.cpp
ArthurSonzogni: Indeed.
(I pushed two patches together, I will fix this part elsewhere)

src/ftxui/component/renderer.cpp
ArthurSonzogni: Thanks!
(I pushed two patches together, I will fix this part elsewhere)

ArthurSonzogni:
Thanks for the review!

I realized it contains another patch that already landed. I will fix your comment in another patch.

PR #178, opened 2021-08-07
Address @tusharpm comments.

ArthurSonzogni:
@tusharpm let two comments in:
https://github.com/ArthurSonzogni/FTXUI/pull/175
but this already landed in:
https://github.com/ArthurSonzogni/FTXUI/pull/173

Address them here.


tusharpm (approved):
These changes look good! Thanks for addressing the comments.

Issue #153, opened 2021-07-14
Drop support for wstring and use utf-8 only?


cmorganBE:
Thoughts on dropping wstring and switching to using utf-8 only? It looks like you've got the utf-8 parser in the code. The use of wstring is interesting, I haven't seen it before, but I'm not sure who its aimed at. The windows unicode stuff is, I thought, deprecated at this point in favor of utf-8. Is there any benefit to using wstring over a normal utf-8 string?

ArthurSonzogni:
Initially, this was for having an easy way to break a string into multiple characters. There is one character per element of the wstring => simple. With a simple std::string, a character might span multiple elements.

Then, it turned out you can have a single character made from multiple wchar_t. That's because of combining characters. Since we are supporting combining characters, then there is not a strong reason to keep working on std::wstring. I just would like to avoid breaking existing users as much as I can.

I momentarily tried that here:
https://github.com/ArthurSonzogni/FTXUI/pull/121#discussion_r654918223

but never when beyond a simple prototype.

cmorganBE:
Ahh gotcha.

I try to see why utf-16 sounded like a good idea but its still unclear to me. utf-8 is such a cleaner approach.

imo the simplest approach is to break existing users vs. trying to introduce wstring->string conversions all over the place. They'd have to drop L everywhere and switch some types but then we'd be utf-8 without using wstring.

BurningEnlightenment:
Have you thought about supporting `char8_t` on the surface API when C++20 is available?

> I try to see why ~utf-16~ utf-32 sounded like a good idea

`sizeof(wchar_t) == 4` holds on most platforms with windows being the notable exception 😉

ArthurSonzogni:
> Have you thought about supporting char8_t on the surface API when C++20 is available?

Requiring C++20 would prevents many people from using the library. Would `char8_t` bring something useful to FTXUI?

I don't believe FTXUI use `char` in its API:
```sh
grep "char " -r include
```
return nothing.

Of course, the API is full of std::string. I think I am going to keep it, since that the most universal string type in C++.

ArthurSonzogni:
@cmorganBE, I addressed this issue.
FTXUI still support std::wstring, but this is deprecated and I will remove them later to let developers time.

ArthurSonzogni:
Fixed.

PR #179, opened 2021-08-08
Prefer std::string over std::wstring.

ArthurSonzogni:
In the past, FTXUI switched from std::string to std::wstring to support
fullwidth characters. The reasons was that fullwidth characters can be
stored inside a single wchar_t.

Then FTXUI added support for combining characters. A single glygh
doesn't even fit a wchar_t. Instead, a glyph can be arbitrary large.

The usage of wstring doesn't really fit the new model and have several
drawbacks:
1. It doesn't simplify the implementation of FTXUI, because of combining
   characters.
2. It reduces drawing performance by 2x.
3. It increase Screen's memory allocation by 2x.

This patch converts FTXUI to use std::string internally. It now exposes
std::string based API. The std::wstring API remains, but is now
deprecated.

Tests and examples haven't been update to show the breakage is limited.
They will be updated in a second set of patches.

Bug: https://github.com/ArthurSonzogni/FTXUI/issues/153


ArthurSonzogni:
+CC @tusharpm  in case you are interested.

This cause FTXUI to use `std::string` instead of `std::wstring`. It makes FTXUI easier to use, improve performance and reduce memory usage.
This is a pretty big change. 
I don't know why, but IWYU decided it was time to properly fix the includes of examples. I am sorry about this.

What do you think about the direction?

include/ftxui/util/ref.hpp
```diff
@@ -59,19 +75,35 @@ class StringRef {
...
+  size_t size() const { return ref_ ? ref_->size() : ref_wide_->size(); }
+  std::string operator[](size_t i) const {
+    return ref_ ? (*ref_)[i] : to_string((*ref_wide_)[i]);
+  }
 
  private:
-  const std::wstring owned_;
-  const std::wstring* address_ = nullptr;
+  const std::vector<std::string>* const ref_ = nullptr;
+  const std::vector<std::wstring>* const ref_wide_ = nullptr;
```
tusharpm, line 106: Consider removing the outer `const` from the data members.
```suggestion
  const std::vector<std::string>* ref_ = nullptr;
  const std::vector<std::wstring>* ref_wide_ = nullptr;
```

Ref: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-constref

src/ftxui/component/event.cpp
```diff
@@ -9,24 +9,23 @@ namespace ftxui {
...
   event.type_ = Type::Character;
   return event;
 }
 
 // static
 Event Event::Character(char c) {
-  return Character(wchar_t(c));
+  std::string input;
+  input += c;
+  return Event::Character(input);
```
tusharpm, line 21:
```suggestion
  return Event::Character(std::string{c});
```

src/ftxui/dom/text.cpp
```diff
@@ -27,80 +28,93 @@ class Text : public Node {
...
-      const int width = wchar_width(c);
-      if (width >= 1) {
-        if (x > box_.x_max)
-          return;
-        screen.PixelAt(x, y).character = c;
-      } else {
-        screen.PixelAt(x - 1, y).character += c;
-      }
-      x += std::max(width, 0);
+    for (auto cell : Utf8ToGlyphs(text_)) {
```
tusharpm, line 31:
```suggestion
    for (const auto& cell : Utf8ToGlyphs(text_)) {
```
For consistency with VText::Render.

src/ftxui/dom/text.cpp
```diff
@@ -27,80 +28,93 @@ class Text : public Node {
...
+  std::string text_;
 };
 
 class VText : public Node {
  public:
-  VText(std::wstring text) : text_(text) {
-    for (auto& c : text_)
-      width_ = std::max(width_, wchar_width(c));
+  VText(std::string text) : text_(text) {
+    width_ = string_width(text_) ? 1 : 0;
```
tusharpm, line 46:
```suggestion
    width_{std::min(string_width(text_), 1)}
```
in the member initializer list.

src/ftxui/dom/text.cpp
```diff
@@ -27,80 +28,93 @@ class Text : public Node {
...
 /// ```bash
 /// Hello world!
 /// ```
-Element text(std::wstring text) {
+Element text(std::string text) {
   return std::make_shared<Text>(text);
 }
 
-/// @brief Display a pieve of unicode text vertically.
+/// @brief Display a piece of UTF8 encoded unicode text.
```
tusharpm, line 91: This overload accepts a `std::wstring` which is not UTF-8 encoded. 

Same for `vtext(std::wstring)` below. 

Also, the doc string examples for these two overloads are missing L"" literals.

src/ftxui/dom/spinner.cpp
```diff
@@ -1,249 +1,249 @@
...
-         L"   /\\O  ",
-         L"    /\\/ ",
-         L"   /\\   ",
-         L"  /  \\  ",
-         L"LOL  LOL",
+         "   /\\O  ",
+         "    /\\/ ",
+         "   /\\   ",
+         "  /  \\  ",
+         "LOL  LO",
```
tusharpm, line 232: nit: replace-all modified this string's content as well.

ArthurSonzogni: Good catch!

include/ftxui/screen/string.hpp
```diff
@@ -12,12 +13,10 @@ std::wstring to_wstring(T s) {
   return to_wstring(std::to_string(s));
 }
 
-int wchar_width(char32_t);
 int wchar_width(wchar_t);
-int wchar_width_cjk(char32_t);
-int wchar_width_cjk(wchar_t);
 int wstring_width(const std::wstring&);
```
tusharpm, line 17: These two are also not used in the library - consider moving to deprecated.hpp or removing.

src/ftxui/dom/text.cpp
tusharpm: Follow-up on the update - the vtext(std::string) overload doesn't mention an encoding.

Also, it might be cleaner to skip `UTF16` for `std::wstring` as *nix systems generally define 4-byte wchar_t.

ArthurSonzogni:
Thanks for the review!

include/ftxui/screen/string.hpp
ArthurSonzogni: Yes, good idea. I will move them to deprecated.

src/ftxui/dom/border.cpp
```diff
@@ -12,10 +13,11 @@
 
 namespace ftxui {
 
-static wchar_t simple_border_charset[] = L"╭╮╰╯─│┬┴┤├";
+static std::string simple_border_charset[] = {"╭", "╮", "╰", "╯", "─",
+                                              "│", "┬", "┴", "┤", "├"};
```
tusharpm, line 17: Consider using `Utf8ToGlyphs` function to split these glyphs from a single raw string literal.

Similarly for charset in gauge and graph.

PR #202, opened 2021-09-10
Border style

ArthurSonzogni:


src/ftxui/dom/separator.cpp
```diff
@@ -52,11 +59,21 @@ class SeparatorWithPixel : public Separator {
 };
 
 Element separator() {
-  return std::make_shared<Separator>();
+  return std::make_shared<Separator>(LIGHT);
 }
 
-Element separator(Pixel pixel) {
-  return std::make_shared<SeparatorWithPixel>(pixel);
```
BurningEnlightenment, line 59: If the removal of the `Element separator(Pixel);` overload was intentional its forward declaration should've also been removed from `elements.hpp`.

BurningEnlightenment (commented):
Can we also add an empty space seperator? I've previously abused `ftxui::separator(ftxui::Pixel{})` to produce this effect:
grafik

ArthurSonzogni:
Yes, I probably removed it by mistake

Le ven. 1 oct. 2021, 20:51, Henrik Gaßmann ***@***.***> a
écrit :

ArthurSonzogni:
@BurningEnlightenment It has been fixed by a recent patch:
https://github.com/ArthurSonzogni/FTXUI/pull/221

BurningEnlightenment:
Thank you for fixing this problem so quickly.

PR #203, opened 2021-09-11
Add an UDL for combined hex RGB colors

BurningEnlightenment:
This is a little helper function I found myself longing for while using the library. It makes defining RGB colors in the common hex format quite convenient:

```c++
auto theme_carbon_grey90() -> theme
{
    using namespace ftxui::literals;
    return theme{.ui_background = 0x262626_rgb,

                 .interactive_01 = 0x0f'62'fe_rgb,
                 .interactive_02 = 0x6f'6f'6f_rgb,
```

In order to allow using the literal on its own it has been put into the
inline namespace `literals`.

### DIFF
--- a/include/ftxui/screen/color.hpp
+++ b/include/ftxui/screen/color.hpp
@@ -1,9 +1,8 @@
 #ifndef FTXUI_SCREEN_COLOR
 #define FTXUI_SCREEN_COLOR
 
-#include <cassert>
 #include <stdint.h>  // for uint8_t
-#include <string>    // for wstring
+#include <string>  // for wstring
 
 #ifdef RGB
 // Workaround for wingdi.h (via Windows.h) defining macros that break things.
@@ -325,17 +324,11 @@ class Color {
 
 inline namespace literals {
 
-/// @brief Creates a color from a combined hex RGB representation, e.g. 0x808000_rgb
-inline Color operator""_rgb(unsigned long long int combined)
-{
-    assert(combined <= 0xffffffU);
-    auto const red = static_cast<uint8_t>(combined >> 16);
-    auto const green = static_cast<uint8_t>(combined >> 8);
-    auto const blue = static_cast<uint8_t>(combined);
-    return Color(red, green, blue);
-}
+/// @brief Creates a color from a combined hex RGB representation, 
+/// e.g. 0x808000_rgb
+Color operator""_rgb(unsigned long long int combined);
 
-}
+}  // namespace literals
 
 }  // namespace ftxui
 
--- a/include/ftxui/screen/color.hpp
+++ b/include/ftxui/screen/color.hpp
@@ -1,6 +1,7 @@
 #ifndef FTXUI_SCREEN_COLOR
 #define FTXUI_SCREEN_COLOR
 
+#include <cassert>
 #include <stdint.h>  // for uint8_t
 #include <string>    // for wstring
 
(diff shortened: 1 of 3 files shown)


include/ftxui/screen/color.hpp
```diff
@@ -1,6 +1,7 @@
 #ifndef FTXUI_SCREEN_COLOR
 #define FTXUI_SCREEN_COLOR
 
+#include <cassert>
```
ArthurSonzogni, line 4: Could you put the implementation and this header inside the .cpp file?

include/ftxui/screen/color.hpp
```diff
@@ -322,6 +323,20 @@ class Color {
...
+    auto const green = static_cast<uint8_t>(combined >> 8);
+    auto const blue = static_cast<uint8_t>(combined);
+    return Color(red, green, blue);
+}
+
+}
+
 }  // namespace ftxui
 
 #endif /* end of include guard: FTXUI_COLOR_H_ */
```
ArthurSonzogni, line 342: Could you add an example in the ./example/dom/ directory?

ArthurSonzogni (commented):
Thanks!
I think I like it. I added some comments below:

include/ftxui/screen/color.hpp
```diff
@@ -322,6 +323,20 @@ class Color {
   uint8_t blue_ = 0;
 };
 
+inline namespace literals {
```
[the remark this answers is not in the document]
BurningEnlightenment: An [inline namespace](https://en.cppreference.com/w/cpp/language/namespace#Inline_namespaces) makes its members also visible in the enclosing namespace. Which in this case means that you can do `using namespace ftxui;` or `using namespace ftxui::literals;` depending on whether you want to make all ftxui symbols visible in the current scope or only the UDL operator. This is considered best practice and the standard library also follows this pattern (s. [chrono-literals](https://en.cppreference.com/w/cpp/symbol_index/chrono_literals), [string-literals](https://en.cppreference.com/w/cpp/symbol_index/string_literals), etc.)

include/ftxui/screen/color.hpp
BurningEnlightenment: done.

include/ftxui/screen/color.hpp
BurningEnlightenment: I amended the `style_color.cpp` example, is this sufficient?

include/ftxui/screen/color.hpp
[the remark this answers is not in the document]
ArthurSonzogni: I just didn't wanted to stay dumb.
I think I figured out why someone would like it. I believe this can be useful if you want to `using namespace ftxui::litterals` and not `using namespace ftxui`.

include/ftxui/screen/color.hpp
```diff
@@ -2,7 +2,7 @@
 #define FTXUI_SCREEN_COLOR
 
 #include <stdint.h>  // for uint8_t
-#include <string>    // for wstring
```
ArthurSonzogni, line 5: Could you bring back the spaces?

ArthurSonzogni (approved):
Thanks!

include/ftxui/screen/color.hpp
ArthurSonzogni: (Ooops, I merged it. I will do it later myself)

BurningEnlightenment: I'm sorry; I completely overlooked that. I hate unnecessary whitespace changes 😓

PR #209, opened 2021-09-20
Added TimeAccumulator in README.md

asari555:
TimeAccumulator was added as a project using FTXUI lib.


ArthurSonzogni:
Really nice use of FTXUI!
Thanks!

PR #222, opened 2021-10-08
do not auto select in menu on focus change

ghost:
In the example `component/menu_multiple` I observed the following issues: 
Using the mouse a hover only changes the focus of a menu -a click changes the selection - that's fine.
But, using the keyboard this was impossible - whatever changed the focus also changed the selection. So I was not free to chose an arbitrary combination via keyboard as I was via mouse - to move to the second menu from the first automatically left me selecting the last entry in the first menu.

Here is my proposed fix:

- arrow keys only change focus
- space changes selection to current focus
- enter remains as it is

### DIFF
--- a/src/ftxui/component/menu.cpp
+++ b/src/ftxui/component/menu.cpp
@@ -56,20 +56,26 @@ class MenuBase : public ComponentBase {
       return OnMouseEvent(event);
 
     if (Focused()) {
-      int old_selected = *selected_;
+      const int old_focused = focused_entry();
+      int new_focused{old_focused};
       if (event == Event::ArrowUp || event == Event::Character('k'))
-        (*selected_)--;
+        --new_focused;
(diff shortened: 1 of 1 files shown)


ArthurSonzogni:
Thanks for this patch! The fundamental difference in between a `Menu` and a `Radiobox` is precisely this behavior. So I think this was working as expected.

Maybe we should provide a second type of `Menu`, based on `Radiobox`, but decorated as a menu?
```cpp
Component Menu2 (ConstStringListRef entries, int* selected) {
  MenuOption option;
  option.style_checked = "> ";
  option.style_unchecked = "   ";
  return Radiobox(std::move(entries), selected, option);
}
```

stephanroslen:
I see, that makes sense, thank you! I guess I can simply add a factory function like that to my source code.

Issue #223, opened 2021-10-08
Question about packaging FTXUI for Gentoo distribution


Petross404:
Hi,

I came across FTXUI and it seems very nice. I want to write an ebuild for Gentoo users to install system-wide this library, but I have two questions:

1) I noticed that compiling with Portage this software, there were static libraries installed
```
>>> Merging sys-libs/ftxui-9999 to /
--- /usr/
--- /usr/lib64/
>>> /usr/lib64/ftxui-screen.a
>>> /usr/lib64/ftxui-dom.a
>>> /usr/lib64/ftxui-component.a
--- /usr/lib64/cmake/
```
Looking at the top-level `CMakeLists.txt` I see that you are importing your targets as static libraries. Can we provide an option for that?

And 2) Your example [here](https://github.com/ArthurSonzogni/ftxui-starter/blob/master/CMakeLists.txt) fetches FTXUI as an external project or something. Is it possible for cmake to find it somehow if it installed in the system?

Thanks!

ezzieyguywuf:
>  I want to write an ebuild for Gentoo

😍 you beat me to it.

Do you plan to publish the ebuild in one of the official gentoo overlays or in your own local overlay?

ezzieyguywuf:
> I recommend developers to point their project to a given git commit of FTXUI, and **let cmake automatically fetch** and compile

This is great for a dev environment or someone that wants to manage their installation locally. However, most packaging systems (I'm most familiar with gentoo's Portage) have some restrictions on network access during compilation/installation - this is a security feature.

Because of that, packaging systems such as Portage need to be able to:

1. Download all necessary artifacts on their own (e.g. using `wget` or something else)
2. Compile the payload (library and/or executable) without accessing network
3. Install payload to a designated location (again, without accessing the network)



> > I see that you are importing your targets as static libraries. Can we provide an option for that?
> 
> I am not sure to understand what is `that`? Do you want dynamic library instead?

Yes, I think he's asking that a cmake option be added for compiling as static or dynamic: if you think this will be challenging, I can take a look at the cmake config and probably submit a PR (this is pretty easy in cmake)

I think a cmake flag for static vs dynamic as well as auto-fetch (your current behaviour) versus "provide your own deps" should probably be all that's needed. Cmake does a pretty good job of "following the rules" regarding installation, so number 3 above shouldn't be an issue.

Petross404:
It seems I am a little late at the party.

[...]

PR #225, opened 2021-10-09
Support dynamic library.

ArthurSonzogni:
- Let the global `BUILD_SHARED_LIBS` dictates whether the library should
  be built statically or dynamically. The cmake's default is statically.
- Add library version and symlink.

This lead to the following install tree.
```
.
├── include
│   └── ftxui
│       ├── component [...]
│       ├── dom [...]
│       ├── screen [...]
│       └── util [...]
└── lib
    ├── cmake
    │   └── ftxui
    │       ├── ftxui-config.cmake
    │       ├── ftxui-config-version.cmake
    │       └── ftxui-config-version-noconfig.cmake
    ├── ftxui-component.so -> ftxui-component.so.0.10.369
    ├── ftxui-component.so.0.10.369
    ├── ftxui-dom.so -> ftxui-dom.so.0.10.369
    ├── ftxui-dom.so.0.10.369
    ├── ftxui-screen.so -> ftxui-screen.so.0.10.369
    └── ftxui-screen.so.0.10.369
```
Fixed: https://github.com/ArthurSonzogni/FTXUI/issues/223

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,19 @@
-# Changelog
+Changelog
+=========
 
-## Current
+Unreleased (development) 
+------------------------
 
 ## Bug
 - On Unix system, fallback to {80,25} screen dimension on failure.
 
-## 0.10 (2021-09-30)
+## CMake
+Added:
+- Support for shared library, via `BUILD_SHARED_LIBS` option.
+- Add library version and symlinks.
+
+0.10 (2021-09-30)
+--------------------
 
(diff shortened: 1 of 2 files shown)


Petross404:
> @Petross404 & @ezzieyguywuf Does that patch works for you? Could you please take a look?

Sure. I will report back.

ArthurSonzogni:
Thanks!
The .so format is what cake did by default. Hopefully this should be right.

PR #258, opened 2021-11-10
added vantage in project section

gokulmaxi:


ArthurSonzogni:
Nice! I will take a closer look, as this probject my be useful to me too ;-)

PR #260, opened 2021-11-16
Fix gridbox segfault

### LINKED ISSUE #259: Segfault when rendering Gridboxes
Hello! I was looking for a TUI library for my project and I've found FTXUI. I've tried it & I love it so far, great job! 
After trying to use a gridbox I've gotten a segfault. This code should reproduce it:

```
gridbox({
         {
             text("1"),
             text("2"),
             text("3"),
             text("4"),
         },
         {},
         {},
```

Upon minor inspection, the root of the issue is in the consturctor of Gridbox, mainly the `filler()` is not inserted to lines due to while loop checking for y_size instead of x_size. Since in code above `x_size` is 4 and `y_size` is 3 it's gonna miss some fillers & later segfault 
on gridbox.cpp:55 in `ComputeRequirement()`. 

I'd be very happy to submit a pull request (my very first open source PR) but if you prefer to do this minor thing yourself that is fine also!

NikolaDucak:
Fixes #259

### DIFF
--- a/src/ftxui/dom/gridbox.cpp
+++ b/src/ftxui/dom/gridbox.cpp
@@ -20,7 +20,7 @@ class GridBox : public Node {
     for (const auto& line : lines_)
       x_size = std::max(x_size, (int)line.size());
     for (auto& line : lines_) {
-      while (line.size() < (size_t)y_size) {
+      while (line.size() < (size_t)x_size) {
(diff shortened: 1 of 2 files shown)


ArthurSonzogni (approved):
Thanks! This is looking good!

PR #274, opened 2021-12-06
adding tabdeeli as a project using ftxui

typon:
Thanks for making this library, it was fun to use. There are some features I want, but overall i could get things working relatively painlessly.


ArthurSonzogni:
Very nice project! I like it a lot!

PR #287, opened 2021-12-23
Feature: Canvas

ArthurSonzogni:
Draw using braille and block characters on a grid.


examples/component/canvas_animated.cpp
```diff
@@ -0,0 +1,257 @@
...
+  int mouse_y = 0;
+
+  // A triangle following the mouse, using braille characters.
+  auto renderer_line_braille = Renderer([&] {
+    auto c = Canvas(100, 100);
+    c.DrawText(0,0, "Several lines (braille)");
+    c.DrawPointLine(mouse_x, mouse_y, 80, 10, Color::Red);
+    c.DrawPointLine(80, 10, 80, 40, Color::Blue);
+    c.DrawPointLine(80, 40, mouse_x, mouse_y, Color::Green);
+    return ElementFrom(std::move(c));
```
felixjulianheitmann, line 25: The amount of features given through the Canvas member functions is great. It also looks really intuitive, good work!
However, as I said in my comment in the feature request, with `Canvas` a user has to manually transform the canvas object into a renderable element which is a different behaviour with respect to all other created elements.

examples/dom/flexbox.cpp
```diff
@@ -0,0 +1,38 @@
+#include <stddef.h>                // for size_t
```
felixjulianheitmann, line 1: I think this example doesn't compile yet. It's not included in the tests CMakeLists but if it's unused at the moment, I think it shouldn't be in the master examples.

felixjulianheitmann (commented):
I think the flexbox.cpp example does not belong into the commit. That might be a mistake or I did not get the correct definitions.

PR #288, opened 2021-12-24
Address felixjulianheitmann suggestions.

ArthurSonzogni:
- Remove unused examples/dom/flexbox.cpp
- Add canvas(width, height, void(Canvas&)) method


ArthurSonzogni:
@felixjulianheitmann I added the following methods to produce element from Canvas for FTXUI
```cpp
Element canvas(ConstRef<Canvas>);
Element canvas(int width, int height, std::function<void(Canvas&)>);
Element canvas(std::function<void(Canvas&)>);
```

Thanks for your previous comments. Does this patch looks good to you?

felixjulianheitmann:
Looks good! Now it can be used like any other feature!

Great job 👍

ArthurSonzogni:
Thanks!

PR #291, opened 2021-12-31
Reduce amount of warnings with pedantic compiler

vnepogodin:
-Wsign-conversion


ArthurSonzogni:
Thanks! We probably can add this warning by default if it isn't enabled yet.

PR #292, opened 2022-01-01
improve the window size handle method

mysterywolf:
1. the default window size should be 80x24 rather than 80x25 in VT100.
2. the ioctl return value result should be checked. Some operating systems don't support TIOCGWINSZ this command.


ArthurSonzogni (approved):
This is reasonnable. Thanks for the PR!

I added comments explaining our choices. Since I wasn't able to justify the 80x80 for windows, I switched toward 80x24 too.

mysterywolf:
Thank you

Issue #293, opened 2022-01-02
[Feature Req] A static UI parser based on FTXUI


SAtacker:
Hi, Arthur (not tagging you as you are busy at this moment),
I had an idea to make a quick Static UI tool built on top of FTXUI
Features:
- Hooks to make it feel dynamic
- No runtime dependencies (although we can have FTXUI itself as the only one)
- Quick YAML or JSON like file to generate UI

ArthurSonzogni:
Hello Shreyas!

This might be a lovely idea, depending on what this is at the end. Would you have some ideas about how it could look like?

SAtacker:
Basically the idea is to generate some customisable UI (although static or in future hooks for events). An example could be [Whiptail](https://en.m.wikibooks.org/wiki/Bash_Shell_Scripting/Whiptail).

ArthurSonzogni:
So your goal is to avoid compiling C++, but let user's to define some textual UI themselve. The UI would be interpreted.

That's a good thing to have, but the difficulty is to make a language that is nice to use and sufficiently expressive.

SAtacker:
Yes I agree, it's difficult and even more so with event based/ hooks.
We need concepts of Signals for events, State of component and communication between events I guess (context: Qt).

omeid:
Sounds like HTML to me.

SAtacker:
https://github.com/SAtacker/quick-ftxui ongoing, maybe it will complete in Summers next year

PR #297, opened 2022-01-05
Constexpr support

vnepogodin:
It's possible to mark some functions as `constexpr`.
Which leads to some memory changes and slight performance improvements

### DIFF
--- a/include/ftxui/screen/screen.hpp
+++ b/include/ftxui/screen/screen.hpp
@@ -62,8 +62,8 @@ class Screen {
   void Print();
 
   // Get screen dimensions.
-  int dimx() { return dimx_; }
-  int dimy() { return dimy_; }
+  int dimx() const { return dimx_; }
+  int dimy() const { return dimy_; }
 
   // Move the terminal cursor n-lines up with n = dimy().
   std::string ResetPosition(bool clear = false);
--- a/include/ftxui/util/ref.hpp
+++ b/include/ftxui/util/ref.hpp
@@ -13,9 +13,9 @@ class ConstRef {
   ConstRef() {}
   ConstRef(T t) : owned_(t) {}
   ConstRef(const T* t) : address_(t) {}
-  const T& operator*() { return address_ ? *address_ : owned_; }
-  const T& operator()() { return address_ ? *address_ : owned_; }
-  const T* operator->() { return address_ ? address_ : &owned_; }
+  const T& operator*() const { return address_ ? *address_ : owned_; }
+  const T& operator()() const { return address_ ? *address_ : owned_; }
+  const T* operator->() const { return address_ ? address_ : &owned_; }
 
  private:
   T owned_;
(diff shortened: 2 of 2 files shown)


ArthurSonzogni:
Thanks you!

I think I have trouble understanding the benefits of using noexcept, constexpr and various things in this patch. They cost some increase in length of the headers, I like to keep them minimal and quick to read. Because more keywords are used, the signal/noise ratio decrease.

I think I am going to cherry-pick the ones I like. I might also take the others if I can be convinced they are worth it. I will take a look tonight. Thanks!

vnepogodin:
@ArthurSonzogni , thanks for the response!

Yeah `constexpr` needs to be in the headers
So it can be evaluated at compile-time
You can also take a look at constexpr micro-kernel with constexpr testsuite(https://www.youtube.com/watch?v=OcyAmlTZfgg)

`noexcept` means that function will not throw an exception
`noexept(false)` means opposite
It is good for api

Also there is [[nodiscard]] which basically prevents user to use return value of the function

ArthurSonzogni:
Yes, I already know noexcept and constexpr. I was wondering why using them worth it?

ArthurSonzogni:
Thanks again for the time you took!

I read it carefully. I think I will take:
- Add const to class attributes when missing.
- Add const to class functions when missing

I think I won't take:
- Add constexpr: I don't understand how it is beneficial. This makes the API stricter, by providing the guarantee it must always be possible to execute it at compile time. This would prevent me from changing the implementation without creating an API breaking change. There are a few cases, where I would don't mind, but I would prefer not to care about this C++ feature.
- Add noexcept: C++ exceptions are [banned](https://google.github.io/styleguide/cppguide.html#noexcept) in FTXUI. Virtually everything could be declared as noexcept. I don't want to add such noise everywhere. I took a look at Chrome and its content/ directory. It seems [noexcept isn't in use and was mostly introduced mistakenly.](https://source.chromium.org/search?q=noexcept%20file:content%20-file:out%20-test%20-third_party&ss=chromium)

Things I am not sure yet:
- Use aggregated initialization instead of assignment initialization. I don't really like the syntax, but this should be mostly a no-op in this case.

I will modify this PR and land it. Should I proceed?

vnepogodin:
@ArthurSonzogni, Yeah sure.

I don't follow Chrome. I just watch at GCC, Clang or even Firefox codebase.

I will anyway try to make it more in use of C++ features, but not in the upstream FTXUI.

ArthurSonzogni:
> I will anyway try to make it more in use of C++ features, but not in the upstream FTXUI.

If you want to maintain a diverging FTXUI project, I guess the constexpr/noexcept have to be useful to you. Out of curiosity, what are your use cases? If I understand them, maybe they might be useful to somebody else and could be part of FTXUI?

vnepogodin:
I'm doing CLI installer for CachyOS.
I should make it public at Github next week.

And I want to make it capable of running installer in just TTY. As I saw it does not work properly in a TTY(became slow, fonts are missing).

If you want to contribute in a future. It will be much appreciated!

And thanks @ArthurSonzogni for your great project

vnepogodin:
> I'm doing CLI installer for CachyOS.
> I should make it public at Github next week.
> 
> And I want to make it capable of running installer in just TTY. As I saw it does not work properly in a TTY(became slow, fonts are missing).
> 
> If you want to contribute in a future. It will be much appreciated!
> 
> And thanks @ArthurSonzogni for your great project

It can help that ticket at some point
https://github.com/ArthurSonzogni/FTXUI/issues/293

Issue #296, opened 2022-01-03
Forcing the selection of a menu item.


azchatlanin:
Hello. I created a menu. Then at some point in time, I need to change this menu. And the size of the menu becomes smaller than it was.

If the selected menu item was larger than the size of the new menu before the change, then there is no selection and the menu item becomes indeterminate until you move the mouse or keyboard:)

Is there any official way to change for example focused_entry() = 0?
I couldn't get to him. Or I'm doing something wrong.

Example:
```
#include <string>
#include <vector>
#include <memory>
#include <filesystem>
#include <algorithm>
#include <iterator>

#include <ftxui/component/captured_mouse.hpp>
#include <ftxui/component/component.hpp>    
#include <ftxui/component/component_base.hpp>
#include <ftxui/component/component_options.hpp> 
#include <ftxui/component/screen_interactive.hpp>
#include <ftxui/dom/elements.hpp>

int main(int argc, const char* argv[]) {

  std::vector<std::string> current_list = { "asdf 1", "adsf 2", "adsf 3", "asdf 4" };

  int selected = 0;
  auto files = ftxui::Menu(&current_list, &selected);

  auto files_rnd = ftxui::Renderer(files, [&] {
    return ftxui::hbox({
      ftxui::vbox({
        ftxui::text("some text"),
      }) | ftxui::border | size(ftxui::WIDTH, ftxui::EQUAL, 800) | ftxui::xflex,

      ftxui::vbox({
        files->Render(),
      }) | ftxui::border | size(ftxui::WIDTH, ftxui::EQUAL, 800) | ftxui::xflex,

      ftxui::vbox({
        ftxui::text("some text"),
      }) | ftxui::border | size(ftxui::WIDTH, ftxui::EQUAL, 800) | ftxui::xflex,
    }) | ftxui::flex;
  });

  auto container = ftxui::Container::Vertical({
    files_rnd,
  });

  auto main_container = ftxui::CatchEvent(container, [&](ftxui::Event event) {
    if (event == ftxui::Event::Character('h')) 
    {
      current_list.erase(current_list.begin());
      current_list.erase(current_list.begin());
      return true;
    }

    return false;
  });


  auto screen = ftxui::ScreenInteractive::Fullscreen();
  screen.Loop(main_container);
}
```
https://user-images.githubusercontent.com/21025345/147914105-f3e6f174-1705-4963-acf3-af650a64f8b6.mp4

ArthurSonzogni:
[...]

PR #298, opened 2022-01-06
Forcing menu

azchatlanin:
fix: https://github.com/ArthurSonzogni/FTXUI/issues/296#issue-1092343846
I don't think these changes should be in RadioboxBase and ToggleBase. Because when rebuilding them, the user still has to make a choice again. I didn't need them yet, but I'll work with them, if I have any ideas, I'll offer something.


ArthurSonzogni:
Thanks!

Could you please take a look at the error?
I still think we want this change for RadioboxBase and ToggleBase too. If you want, you can add them.

ArthurSonzogni:
I added some tests and tweaked the implementation a bit.

Thanks again!

PR #300, opened 2022-01-09
Fix forcing

azchatlanin:
I think that after rebuilding the menu and other similar components, the menu position should be set to zero. Otherwise, there may be a situation when I reached the middle of the menu and rebuilt it, and the new rebuilt menu became no smaller than the previous one. And then it won't be much fun to go back to the beginning. Example of nnn or ranger (file navigation)


ArthurSonzogni:
Thanks!
I don't immediately see the difference with:
https://github.com/ArthurSonzogni/FTXUI/commit/358f886fabe3486d52533d891aa2d0105e4845b2

Could you please rebase on top, so that we can see it in the diff?

azchatlanin:
For my commit [eb09b49](https://github.com/ArthurSonzogni/FTXUI/pull/300/commits/eb09b495c8f36ce6d69a4800613b05154cd0f6bc) I see - 1 workflow awaiting approval.

azchatlanin:
I fixed it аor my commit eb09b49 I see - 1 workflow awaiting approval.

azchatlanin:
In fact, it's all plugging holes. We should have a method of forcing the selected item to be set depending on the event. I.e. if I pressed a key, then I can set the selection I need for this event.

ArthurSonzogni:
> We should have a method

The selected element and the focused element are both provided externally. So I don't see any strong reason to believe the user can't modify those values on events:

We already have. However it is not easily discoverable, but I don't believe it can interest most users:
```cpp
  std::vector<std::string> entries = {"1", "2", "3"};
  int focused_entry = 0;
  int selected_entry = 0;
  MenuOption option;
  option.focused_entry = &focused_entry;
  auto menu = Menu(&entries, &selected, option);
```

Does knowing this resolve your problem?

azchatlanin:
> 
> ```c++
>   option.focused_entry = &focused_entry;
> ```

Yeah, cool, I didn't see that!

azchatlanin:
It is necessary to reject this PR in this case it does not make sense.

PR #305, opened 2022-01-12
Fix README link

pezy:
- [Build using nxxm] is same with [Build using CMake], just remove
- fix [Build using CMake] link
- add basic .gitignore file


ArthurSonzogni:
Thanks! This is looking good. Could you please remove the two unrelated change ".gitignore" and "Menu.cpp"  before I merge the PR?

PR #306, opened 2022-01-12
Fix Clamp crash when entries_ size is zero

pezy:
Run ftxui_example_homescreen on Windows, then select compiler tab, crash on the origin code.


ArthurSonzogni:
Thanks!
I can't reproduce. Could you please let me know the location where this is crashing?

pezy:
1. Run the  ftxui_example_homescreen.exe in Visual Studio 2019 Debug mode, then select compiler tab, there is a assertion failed:
image

2. Retry to location the problem:
image

std::clamp need `_Min_val` less than `_Max_val`.

3. in menu.cpp:141

```cpp
  void Clamp() {
    boxes_.resize(size());
    *selected_ = std::clamp(*selected_, 0, size() - 1); // !!! wow, size() = 0 --> clamp(*selected_, 0, -1)
    focused_entry() = std::clamp(focused_entry(), 0, size() - 1);
  }
```

ArthurSonzogni:
Thanks I see!

I think I would prefer redefining std::clamp by:
```cpp
template<class T>
constexpr const T& clamp( const T& v, const T& lo, const T& hi)
{
    return v < lo ? lo : hi < v ? hi : v;
}
```

I will update your patch tonight and merge it.

ArthurSonzogni:
Thanks!

Issue #301, opened 2022-01-09
[Question] Is it possible to "suspend" FTXUI component from writting to stdout


vnepogodin:
Is it possible to " suspend"  FTXUI component from writting to stdout.
And " resume"  after child process is done.

For example:
I want to execute `bash` interactively(so I need to block FTXUI from writting to the input stream),
and after `exit`, unlock FTXUI

ArthurSonzogni:
Yes kind of...

I would do:
```cpp

auto screen = ScreenInteractive::Fullscreen()
auto exit_loop = screen.ExitClosure(); // Movable and thread safe function.

auto component = CreateMyComponent(exit_loop);

while(true) {
  screen.Loop();
  execute_bash_interactively()
}
```

And someone somewhere call:
```cpp
exit_loop();
```


Alternatively, it could be possible to expose and call:
https://github.com/ArthurSonzogni/FTXUI/blob/84299de2e11522ad1d531a50f7610678ae0e5e67/include/ftxui/component/screen_interactive.hpp#L35-L36
while inside the loop, but I am not sure without giving it more thoughts.

vnepogodin:
Thanks!
I'll expose those functions. And try if it helps.

exit_loop will not help me as I spawn component from another component and if I execute `execute_bash_interactively` it will just ignore it

vnepogodin:
@ArthurSonzogn,  thanks again!
It worked.
Should I create PR with example?

ArthurSonzogni:
Great! Yes, you can submit a PR. I will have to work a bit on top of your change if I want to merge it. I will add example.

vnepogodin:
> Great! Yes, you can submit a PR. I will have to work a bit on top of your change if I want to merge it. I will add example.

Ok I did submit a PR with example.

PR #307, opened 2022-01-12
[Feature]  "Suspend" and "Resume" the screen

vnepogodin:
🚧 suspend FTXUI on interactive actions such as "bash" shell

Resolves: https://github.com/ArthurSonzogni/FTXUI/issues/301


vnepogodin:
so I did resolve error from Windows CI
Which was related to c++20 feature:
member initializer list.
https://en.cppreference.com/w/cpp/language/constructor

ArthurSonzogni:
Thanks!
Just a quick note: I am a bit busy at the moment. I will try to take a look very soon.

ArthurSonzogni:
Thanks! I took a look. I applied the following changes:
- Simplify the example. Keep only the essential. Goal is to keep examples short to favorise developers to read them.
- Instead of exposing Suspend() / Resume(), I preferred exposing a callback decorator. This ensures for every "Uninstall()", there is a matching "Install()" called.
- Remove the parts about clearing the screen. I was not really understanding its purpose and doesn't look correct clearing the screen after uninstalling it. I was not sure the command was correct. Developers can anyway clear the screen themselves if they want. I prefer letting them the choice.

Does it looks good to you?

vnepogodin:
Yeah looks good. I'm ok with callback function instead of exposing already existed.
Example looks good too. It's simple though as it should be

I was clearing screen because then I run`bash`(for example) and drag mouse it starts outputting weird characters (you can confirm that from example I created later on)

ArthurSonzogni:
> I was clearing screen because then I runbash(for example) and drag my mouse it starts outputting weird characters to stdout stream(you can confirm that from example I created later on)

Good observation. We were forgetting to call std::flush in order to commit the new state before proceeding further. This has now been fixed by the latest patch.

PR #321, opened 2022-02-01
Replace std::clamp with util::clamp and reformat the code

NikolaDucak:
I've gotten a build error complaining about there being no `clamp` identifier in `std` namespace.
Snooping around showed me that `std::clamp` caused some issues and was replaced. These two lines were left out, this should fix my build issues and clean up the removal of `std::clamp`.
Also, `clang-format`-ed the source.


ArthurSonzogni:
Thanks!

PR #325, opened 2022-02-05
Feature: Vertical gauge

brakmic-aleksandar:
Hey, i've implemented vertical gauge, it could be useful for process monitors, equalizers etc.
Some things that i could improve here are: 
- [x] Rename members of `enum GaugeDirection { HORIZONTAL, VERTICAL };` into something more specific to avoid future name clashes or i could just rename GaugeDirection into Direction and make it general enum
- [x] Implement remaining two directions
-  [ ] Add example code
-  [ ] Microsoft fallback charset is currently same as normal one, i only have Linux and macOS machines, so i can't test whether they render properly

Edit: Added checklist for possible improvements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
 Changelog
 =========
 
-unreleased (development) 
-------------------------
+current (development) 
+---------------------
+
+2.0.0
+-----
 
 ### Features:
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.11)
 
 project(ftxui
   LANGUAGES CXX
-  VERSION 0.11.1
+  VERSION 2.0.0
 )
 
 option(FTXUI_BUILD_DOCS "Set to ON to build docs" ON)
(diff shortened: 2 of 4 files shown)


ArthurSonzogni:
You are awesome! Thanks!
I wanted this feature for a long time, but didn't took time to make it.

The code looks very clean! I am going to take a look.

I am a bit worried introducing an enum that I might want to change later and cause an API break. I believe we also would like at some point to have gauge going into the 4 directions (left, up, right, down).

At first glance, I believe I would like to change a bit your patch, to allow (in the future) an API like this one:
```cpp
enum class Direction { Left, Up, Right, Down };
Element gauge();
Element gaugeLeft();
Element gaugeUp();
Element gaugeRight();
Element gaugeDown();
Element gaugeDirection(Direction);
```

brakmic-aleksandar:
I've renamed them to Up/Right, i couldn't rename the enum because name Direction is already taken. 

Also i wouldn't mind implementing the other two as well, from what i see [UTF-8 selection of RIGHT and UP blocks isn't as wide as DOWN and LEFT](https://www.utf8-chartable.de/unicode-utf8-table.pl?start=9472#:~:text=ONE%20EIGHTH%20BLOCK-,u%2B2590,-%E2%96%90) so that might make progression a bit choppier.

brakmic-aleksandar:
I've implemented all directions. :)

ArthurSonzogni:
Awesome!
Please let me just tweak it a bit, before landing this patch.

ArthurSonzogni:
Ooops, I messed up with git. I wasn't able to push on your branch. Then I followed some git instruction and I ended up pushing my master toward yours instead :-(

I am going to push and open a new PR instead. Sorry.

brakmic-aleksandar:
That's fine :). I would be lying if i said similar thing hasn't happened to me in the past.

PR #326, opened 2022-02-06
Gauge direction

ArthurSonzogni:
Originally, from: https://github.com/ArthurSonzogni/FTXUI/pull/325
By: [brakmic-aleksandar](https://github.com/brakmic-aleksandar)


ArthurSonzogni:
@brakmic-aleksandar 
Could you please take a look?

I applied the following changes:
- Run `./tools/iwyu.sh` (which also run `./tools/format.sh`)
- Add tests. This was useful, because I had to fix the implementation for `ratio = 0` and `ratio = 1.f`.
- I notices the unicode characters aren't as detailed for one direction than the other one. We are missing some frames. I decided to draw one direction and the opposite the same way, but with a complementary ratio and an inverted style. This way, we only have to implement right and down and get up and left for free, with a better quality.
- Make `inverted` to toggle the inverted bit. This way, the `inverted` decorator always do something.
- Add entries to the CHANGELOG.

brakmic-aleksandar:
Looks great! Good idea to use `inverted` decorator!

ArthurSonzogni:
Merged!
Thank you so much for this feature!

Issue #330, opened 2022-02-10
Add support for SIGSTP


ArthurSonzogni:
### Discussed in https://github.com/ArthurSonzogni/FTXUI/discussions/309

<div type='discussions-op-text'>

<sup>Originally posted by **MattBystrin** January 12, 2022</sup>
Hi, @ArthurSonzogni ! 
Today I've noticed that ScreenInteracitve do not works with SIGSTP. I've rarely use ctrl+z in term in case i usualy work with tmux, but maybe someone finds it handy. What do you think ?
</div>

MattBystrin:
Sorry, I misspelled the signal name. The right version is SIGTSTP

ArthurSonzogni:
Fixed.

MattBystrin:
Hello @ArthurSonzogni ! 
I've checked the improvements - everything is working as expected. 
Great job !

ArthurSonzogni:
Thanks!

PR #331, opened 2022-02-12
Support SIGTSTP and task posting.

ArthurSonzogni:
- Add support for SIGTSTP:
https://github.com/ArthurSonzogni/FTXUI/issues/330
This

- Add support for task posting.
This allows folks to defer function execution, and execute it directly
below the main loop. The task are executed in a FIFO order.


ArthurSonzogni:
@MattBystrin FYI.
Here is the patch adding support for SIGSTP.

PR #340, opened 2022-02-18
don't override PREFIX target property

ericLemanissier:
PREFIX is by default "lib", which most of the libraries use.
this changes makes the library files libftxui-foo.a
as ArthurSonzogni/FTXUI#140 intended originally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,12 @@ Changelog
 current (development) 
 ---------------------
 
-#### DOM:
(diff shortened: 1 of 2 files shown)


ericLemanissier:
AFAIU, this issue is fixed because the libraries are in the lib folder now

ArthurSonzogni:
Thanks! This sounds good to me!
I added this entry to the CHANGELOG. This is a breaking change. This will be released in 3.0.0.

PR #354, opened 2022-03-10
Component decorators

tchaikov:


include/ftxui/component/component.hpp
```diff
@@ -66,10 +66,20 @@ Component Renderer(std::function<Element()>);
...
                       Component child,
                       Ref<bool> show = false);
 
+// -- Decorator --
+template<typename Decorator>
+Component operator|(Component component, Decorator&& decorator) {
+  return decorator(std::move(component));
+}
+using ComponentDecorator = std::function<Component(Component)>;
+ComponentDecorator CatchEvent(std::function<bool(Event)>&& on_event);
```
tchaikov, line 80: @ArthurSonzogni can we use lower-case name for these decorators like what we have in `include/ftxui/dom/elements.hpp` ?  for two reasons

1. more consistent this way
2. avoid the name collisions -- so we can have a `renderer()` decorator.

ArthurSonzogni: I am a bit sad about the different style being used, I agree. However using both `Renderer` and `renderer` would be even worse.
I don't believe there is any problem for the `renderer` component. I just added it.

ArthurSonzogni:
Thanks!

I added the `Renderer` decorator. The |= operator.
I also added ways to decorate a component, with an Element decorator.

Could you please take a look?

tchaikov (commented):
lgtm in general

include/ftxui/component/component.hpp
```diff
@@ -38,38 +47,55 @@ Component Tab(Components children, int* selector);
...
+
 Component ResizableSplitLeft(Component main, Component back, int* main_size);
 Component ResizableSplitRight(Component main, Component back, int* main_size);
 Component ResizableSplitTop(Component main, Component back, int* main_size);
 Component ResizableSplitBottom(Component main, Component back, int* main_size);
+
 Component Renderer(Component child, std::function<Element()>);
 Component Renderer(std::function<Element()>);
 Component Renderer(std::function<Element(bool /* focused */)>);
+ComponentDecorator Renderer(ElementDecorator);
```
tchaikov, line 86: i see, i was trying to implement
```c++
ComponentDecorator Renderer(std::function<Element()>);
```

include/ftxui/component/component.hpp
tchaikov: okay. commented inlined in the code.

tchaikov:
@ArthurSonzogni modulo the nit, lgtm. shall i squash the commits or you will do so?

src/ftxui/component/renderer.cpp
```diff
@@ -104,6 +104,28 @@ Component Renderer(std::function<Element(bool)> render) {
   return Make<Impl>(std::move(render));
 }
 
+/// @brief Decorate a component, by decorating what it renders.
+/// @param decoratorr the function modifying the element it renders.
```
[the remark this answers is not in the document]
ArthurSonzogni: Thanks!

ArthurSonzogni:
Merged!
Thanks for your contribution, I like the idea a lot!

Issue #370, opened 2022-04-03
Should Buttons work in a Frame?


Serinox:
Trying to learn how to use the library and looking through all of the examples. I tried to change the checkbox_in_frame.cpp example code to use Buttons in the frame instead of checkboxes. When I get is a frame that doesn't scroll down with the selected item when I navigate through the buttons. 

Should this work or have I missed something?

If it helps any I'm writing this on MacOS 12.3.1 using CLion and Apple clang version 13.1.6 (clang-1316.0.21.2) and below is a minimal modified main method code that reproduces the behavior.

```auto container = Container::Vertical({});
for (int i = 0; i < 30; ++i) 
{
    container->Add(Button("Button #" + std::to_string(i), [&]()->void{},ButtonOption::Ascii()));
}
auto renderer = Renderer(container, [&] {
        return container->Render() | vscroll_indicator | frame |
            size(HEIGHT, LESS_THAN, 10) | border;
    });
auto screen = ScreenInteractive::FitComponent();
screen.Loop(renderer);
return 0;
```

ArthurSonzogni:
You are right. It is missing.
I will fix this.

ArthurSonzogni:
Hello @Serinox
Could you please take a look?

The difficulty is that we used to "TakeFocus()" when hovering the button with the mouse. This makes it very difficult to deals with it in a frame, because it would make view to scroll down toward the end, when the mouse over the next button.

I had to "TakeFocus()" only when clicking.  This cause two button to appear as focused, the one focused by the mouse and the one focused by the keyboard.

Serinox:
Oh wow that was a fast reply! 

I can confirm that the branch you pushed does seem to fix the problem! Many thanks!

PR #371, opened 2022-04-03
Make the focused button to take the focus inside frame.

ArthurSonzogni:
This resolves:
https://github.com/ArthurSonzogni/FTXUI/issues/370


ArthurSonzogni:
Hello @Serinox
Could you please take a look?

The difficulty is that we used to "TakeFocus()" when hovering the button with the mouse. This makes it very difficult to deals with it in a frame, because it would make view to scroll down toward the end, when the mouse over the next button.

I had to "TakeFocus()" only when clicking. This cause two button to appear as focused, the one focused by the mouse and the one focused by the keyboard.

PR #70, opened 2021-02-08
Work around Windows.h breaking the Color::RGB function

robinlinden:
wingdi.h (included via Windows.h) defines an RGB macro that breaks
things. If a user really wants that macro in the same file as FTXUI they
can move the Windows.h include to after the inclusion of FTXUI's
headers.

Fixes #59

PR #394, opened 2022-05-04
windows.h defines DrawText and shadows the function Canvas.DrawText, …

SColibri:
Fix based on solution #70.


ArthurSonzogni (approved):
Thank you!

PR #403, opened 2022-05-21
Fix PostEvent() segfault

XNUConner:
Fix segfault when PostEvent() called on inactive screen.


ArthurSonzogni:
Thanks!
I added tests and extended the scope of the fix toward Task as well.

PR #406, opened 2022-05-23
ci: use gcovr 5.0 temporarily

aminya:
Related to https://github.com/aminya/setup-cpp/issues/86


ArthurSonzogni:
Thank you so much for taking the time to fix this!
Let's see what happens with the CQ and I will merge this patch.

PR #409, opened 2022-06-01
Swap incorrect width/height mapping

jhon:
width and height were being set using the incorrect axes resulting in incorrect canvas dimensions


ArthurSonzogni:
Thanks! You are right.
We just need to update the gold results. See test failure. Sorry about this, this isn't the be kind of tests.

I will do it later today for you if you don't have the time.

ArthurSonzogni:
Thanks!

PR #415, opened 2022-06-08
Set includes as system interface to suppress warnings for other users

vebjornjr:
Sets the include directories as `SYSTEM INTERFACE` for users of this library to suppress compiler warnings in these header files (from e.g. -Weffc++).

To ensure that compiler warnings always appear when the project is opened stand-alone, I included a small check on the project name. This is not a perfect solution. If users of this library use FetchContent before setting their own project name, they will still get compiler warnings. However, I think it's an improvement.

I checked that the compiler warnings disappear in my other project when I FetchContent my fork, and that I still get compiler warnings when opening the fork as a stand-alone project.

### DIFF
--- a/cmake/ftxui_set_options.cmake
+++ b/cmake/ftxui_set_options.cmake
@@ -13,9 +13,11 @@ function(ftxui_set_options library)
     set_target_properties(${library}
       PROPERTIES CXX_CLANG_TIDY "${CLANG_TIDY_EXE};-warnings-as-errors=*"
     )
-  endif()
 
-  if(${CMAKE_PROJECT_NAME} STREQUAL ${PROJECT_NAME}) # check if project is stand-alone
+    # By using "PUBLIC" as opposed to "SYSTEM INTERFACE", the compiler warning
+    # are enforced on the headers. This is behind "FTXUI_CLANG_TIDY", so that it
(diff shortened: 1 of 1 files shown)


vebjornjr:
There exists a `PROJECT_IS_TOP_LEVEL` variable in CMake, but it was introduced in version 3.21. So I skipped it since Ubuntu 20.04 is stuck at version 3.16 (through apt).

ArthurSonzogni:
Thanks! I took some times to try myself and check if everything was working as intended.

I made some tweaks to avoid checking for the project name.
Does this change looks good to you?

vebjornjr:
Nice tweak on the project name.

However, I don't think the `SYSTEM` works the same way as `PUBLIC`/`PRIVATE`. I think it affects the whole command, which causes the src/ folder to be included as a system header (-isystem on the command line). When I opened the project (with your commit) and set the FXTUI_CLANG_TIDY option it seemed to confirm this. It doesn't affect me as a user of this library, but I guess you want it to be included the normal way (with warnings). So I think the `SYSTEM INTERFACE` and `PRIVATE` must be in two different commands. I also moved the `SYSTEM` keyword to emphasize that it does work differently.

ArthurSonzogni:
Perfect!

ArthurSonzogni:
Thank you!

Issue #290, opened 2021-12-29
Probably reduce memory usage


vnepogodin:
example "homescreen" uses around ~150MB of virtual memory
example "gallery" uses around ~79MB of virtual memory

I wonder is it possible to reduce virt memory usage?

ArthurSonzogni:
It sounds like way too much! How big is your terminal dimensions? Are you on Linux?
I will take a look and verify how much is needed.

ArthurSonzogni:
I took a look. Note that this is virtual memory, not residential memory. This means the program has some address space in its memory map, but it isn't used, and it doesn't take any space out of the RAM.

I tried "homescreen" and it only uses 4.7 Mo (4706 Ko).
However its virtual one is indeed 151MB as you said

![image](https://user-images.githubusercontent.com/4759106/147710418-bf5b57a6-e408-42a5-abeb-3e0bebf9d626.png)

I admit I don't know why the virtual memory is high compared to the real memory used. Do you have ideas? Would you know why we should care?

vnepogodin:
> It sounds like way too much! How big is your terminal dimensions? Are you on Linux? I will take a look and verify how much is needed.

was 136 columns then I was testing. Using Arch Linux

vnepogodin:
btw I made it faster a bit, but memory usage is changed
![image](https://user-images.githubusercontent.com/37552959/147819859-e637e40d-014d-48a9-b956-a124d8a44a7a.png)

ArthurSonzogni:
@vnepogodin Thanks for taking a look.
If you manage the find the root cause of this virtual memory usage, please let me know.
I suspect this is the way the allocator works. It request a lot of addressable space, without committing any data to it.

vnepogodin:
Update:
Memory usage is changed after introducing animation and other cool features
I think that virtual memory usage is high because of how `shared_ptr`s works :eyes: 
![image](https://user-images.githubusercontent.com/37552959/158251069-993033bb-2c38-4fd4-88bf-1dfacfe732a8.png)

ArthurSonzogni:
Thanks for checking it!

I honestly don't have any idea what is going on. If the problem was about some `shared_ptr` cycle unable to deallocate, we would have seen the residential memory to increase.

I checked running valgrind:
```
==7396== HEAP SUMMARY:
==7396==     in use at exit: 0 bytes in 0 blocks
[...]

PR #420, opened 2022-06-13
Avoid making new allocation to clear the screen.

ArthurSonzogni:
Previously, a new 2D vector was allocated for every new frame. This
caused a lot of temporary allocation to be made.

This patch modify "Screen::Clear" so that it do make a new allocation,
but clear the existing one instead.

Bug:https://github.com/ArthurSonzogni/FTXUI/issues/290#issuecomment-1153327251


ArthurSonzogni:
Hello @vnepogodin,
Could you please take a look?

PR #421, opened 2022-06-14
Fix border charset array being to large

AMS21:
This for some reason caused the clang compiler to crash, while also
being incorrect as the tables are actually only 5x6.

See the LLVM issue here:
https://github.com/llvm/llvm-project/issues/56016

This patch was suggested by @tbaederr


LunarWatcher:
Please consider making a release for this as well. At the very least a patch tag for the commit fixing the bug.

Both the GH and Conan 3.0.0 releases are broken, it's still the newest version, and the version that's recommended without any mention of Clang 14/15 misbehaving

Issue #430, opened 2022-07-04
Thread-safety issues in an example


dedlocc:
I suspect the "homescreen" example to be non-thread-safe. `refresh_ui_continue` is being written to by one thread and (potentially) simultaneously read from another, while not being atomic.

https://github.com/ArthurSonzogni/FTXUI/blob/0abaab626859efbe5e45b63cd6d2db5312396cbf/examples/component/homescreen.cpp#L502-L513

ArthurSonzogni:
That's true. There are no consequence here, but still. It would be nice showing some considerations and post a task to be executed in the UI thread. I will make an update.

PR #431, opened 2022-07-04
Fix homescreen example thread safety.

ArthurSonzogni:
This addresses and fix:
https://github.com/ArthurSonzogni/FTXUI/issues/430

This patchs makes |shift| to be read and written on the same thread. We
request the update to be made via a task posted on the main thread.

This patch has no real consequence, the previous behavior was fine.
I hope it will help users not to have thread safety issue and better
understand they can post tasks this way.

### DIFF
--- a/examples/component/homescreen.cpp
+++ b/examples/component/homescreen.cpp
@@ -504,8 +504,13 @@ int main(int argc, const char* argv[]) {
     while (refresh_ui_continue) {
       using namespace std::chrono_literals;
       std::this_thread::sleep_for(0.05s);
-      shift++;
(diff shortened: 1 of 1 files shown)


ArthurSonzogni:
Hello @dedlocc

Could you please take a look?

dedlocc:
`shift` seems good now. Doesn't the issue with `refresh_ui_continue` remain though? There's still a data race (and hence UB).

ArthurSonzogni:
> shift seems good now. Doesn't the issue with refresh_ui_continue remain though? There's still a data race (and hence UB).

That's true. This is now fixed.

PR #442, opened 2022-07-10
Fixes #441

### LINKED ISSUE #441: Canvas::DrawText() draws nothing if the start coordinate is out of bounds
https://github.com/ArthurSonzogni/FTXUI/blob/f91677e79f00dac2a02a08f1c993fc6b9983f867/src/ftxui/dom/canvas.cpp#L804-#L806

I suspect the intent is to draw only those elements of the string that are in bounds, but as soon as any element is not in bounds then x will no longer be incremented. If this occurs on the first element, then the code will iterate through all of the glyphs in the string yet it will draw precisely nothing.

You could solve it by incrementing x when the element is out of bounds as shown here.

```cpp
void Canvas::DrawText(int x,
                      int y,
                      const std::string& value,
                      const Stylizer& style) {
  for (const auto& it : Utf8ToGlyphs(value)) {
    if (!IsIn(x, y)) {
      x += 2; // additional line
      continue;
    }
    Cell& cell = storage_[XY{x / 2, y / 4}];
    cell.type = CellType::kText;
    cell.content.character = it;
    style(cell.content);
    x += 2;
  }
}
```

badlydrawnrod:
Fix for Canvas::DrawText() draws nothing if the start coordinate is out of bounds.

### DIFF
--- a/src/ftxui/dom/canvas.cpp
+++ b/src/ftxui/dom/canvas.cpp
@@ -802,6 +802,7 @@ void Canvas::DrawText(int x,
                       const Stylizer& style) {
   for (const auto& it : Utf8ToGlyphs(value)) {
     if (!IsIn(x, y)) {
+      x += 2;
       continue;
     }
     Cell& cell = storage_[XY{x / 2, y / 4}];


ArthurSonzogni:
Perfect! Thank you for the PR!

ArthurSonzogni (approved):

PR #447, opened 2022-07-29
Typo in iwyu.cmake

jmarrec:
iwyu.imp not iwyu.impl

### DIFF
--- a/cmake/iwyu.cmake
+++ b/cmake/iwyu.cmake
@@ -2,6 +2,6 @@ find_program(iwyu_path NAMES include-what-you-use iwyu)
 if(iwyu_path)
   set_property(TARGET ${lib}
     PROPERTY ${iwyu_path} -Xiwyu
-      --mapping_file ${CMAKE_CURRENT_SOURCE_DIR}/iwyu.impl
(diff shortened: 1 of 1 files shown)


cmake/iwyu.cmake
```diff
@@ -2,6 +2,6 @@ find_program(iwyu_path NAMES include-what-you-use iwyu)
 if(iwyu_path)
   set_property(TARGET ${lib}
     PROPERTY ${iwyu_path} -Xiwyu
-      --mapping_file ${CMAKE_CURRENT_SOURCE_DIR}/iwyu.impl
+      --mapping_file ${CMAKE_CURRENT_SOURCE_DIR}/iwyu.imp
```
jmarrec, line 5: https://github.com/ArthurSonzogni/FTXUI/blob/d9241435ce9fa3d81f0ea5a0d34c3048883629ee/iwyu.imp

ArthurSonzogni: Thanks! I will take a look tomorrow!
This looks good. I only need to check why this wasn't an issue for me.

ArthurSonzogni:
Thanks for your patch!

PR #460, opened 2022-08-19
Fixed dim and bold not mixing well

jansende:
One single reset code controls both the dim and bold properties. Mixing both led to one of the properties being wrongly reset.


ArthurSonzogni:
Thank you!

ArthurSonzogni (approved):
Thanks you!

PR #461, opened 2022-08-19
Fixed divide by zero segault

jessopi:
A divide by zero segfault would occur in vscroll_indicator if terminal window height was shrunk. Added conditional to prevent this from occurring.


ArthurSonzogni (approved):
Thanks!

Issue #453, opened 2022-08-08
The focus moves when an item is selected


ArthurSonzogni:
### Discussed in https://github.com/ArthurSonzogni/FTXUI/discussions/452

<div type='discussions-op-text'>

<sup>Originally posted by **Alexandring** August  8, 2022</sup>
Hello. I am faced with the following problem.
Select an item in the menu with the mouse and the selected item moves to the middle of the window. The same thing happens when I use containers. How do I disable this feature? I want the selected element to remain in its place and scrolling does not happen.</div>

ArthurSonzogni:
Hello @Alexandring,

That's because we put the focus on the selected element:
https://github.com/ArthurSonzogni/FTXUI/blob/master/src/ftxui/component/menu.cpp#L115-L120

That's what you usually want to do when doing a keyboard navigation. This is conflicting with mouse navigation, where you usually don't want to scroll the view when clicking on an element.

Not sure what do do. Maybe we can maintain a "keyboard" and "mouse" index and put the focus only on the keyboard one. That's probably a good task to work on.

Alexandring:
Can I somehow change the behavior of the menu so that it works throughout the application as I want?

ArthurSonzogni:
> @ArthurSonzogni
Can I somehow change the behavior of the menu so that it works throughout the application as I want?

This would require modifying the FTXUI library.
You can try doing the change I explained above, or wait for me to do it.

Alternatively, you can create your own Menu class, behaving the way you would like.

PR #462, opened 2022-08-19
Menu: keep the previously focused element with mouse.

ArthurSonzogni:
This resolves:
https://github.com/ArthurSonzogni/FTXUI/issues/453


ArthurSonzogni:
Hello @Alexandring,
Could you please take a look? Does this resolve your issue?

Issue #467, opened 2022-08-28
Support multiple directions for sliders.


ArthurSonzogni:
### Discussed in https://github.com/ArthurSonzogni/FTXUI/discussions/466

@subhasishk23

<div type='discussions-op-text'>

<sup>Originally posted by **subhasishk23** August 25, 2022</sup>
Is it possible to place Slider component vertically?
The example code provided at examples/component/focus.cpp renders Slider horizontally.
I want to place it vertically to look like a scroller.</div>

Support the 4 directions for the slider element.

PR #468, opened 2022-08-28
Feature: Slider in any directions.

ArthurSonzogni:
Add the `SliderOption` option supporting:
```cpp
{
  Ref<T> value;
  ConstRef<T> min = T(0);
  ConstRef<T> max = T(100);
  ConstRef<T> increment = (max() - min()) / 20;
  GaugeDirection direction = GaugeDirection::Right;
  Color color_active = Color::White;
  Color color_inactive = Color::GrayDark;
};
```

In particular, this supports multiple direction. This resolves:
https://github.com/ArthurSonzogni/FTXUI/issues/467

This one do not support adding a label. The old constructors can still
be used to have a label.


ArthurSonzogni:
Hi @subhasishk23,
Thanks for sponsoring FTXUI!
Do you have any feedback around this patch adding support for multiple directions for the slider Component?

Issue #474, opened 2022-08-31
integration of IO libs that require polling 


niXman:
hi,

for example, `boost::asio`s [io_context](https://www.boost.org/doc/libs/1_80_0/doc/html/boost_asio/reference/io_context.html) can be used as main event loop or as independent event loop.
for the second case simply calling `poll_one()`/`run_one()` periodically is sufficient.

so my question is: how can I implement a periodic function call preferably with the ability to change the frequency?

many thanks!

ArthurSonzogni:
Hello @niXman 

You can run the `ScreenInteractive::Loop()` from one thread.
It will draw new frames only after receiving new events. Events are coming from the terminal. If you need to emit other from different kind of sources, you can submit custom ones in order to produce new frames.
```cpp
screen_interactive::Post(Event::Custom);
```
Post() can send Task or Event. The function is thread safe.

The main loop will return shortly after someone calls the function returned by:
```cpp
screen_interactive::ExitClosure();
```
Calling the returned function is also thread safe.

Does this resolve your issue, or do you want to suggest something more?

niXman:
Hello @ArthurSonzogni 

thanks for your reply!

I think did not understand each other...
does FTXUI provide any timers I can just run and subscribe to its timeout handler/event?


best!

ArthurSonzogni:
If you can explain what you want to achieve, I might be able to provide the right solution.

Some replies:
> subscribe to its event
- Your component can subscribe to [ComponentBase::OnEvent](https://arthursonzogni.github.io/FTXUI/classftxui_1_1ComponentBase.html#ab84629a8fecad2e1cdd426b7ddc72277).
- Alternatively, in a more functional way you wrap your component within [CatchEvent](https://arthursonzogni.github.io/FTXUI/#component-catchevent) function.
> FTXUI provide any timers

I would suggest you spawn a thread and make it loop with a timer.

niXman:
hello again!

just for explanation:
```
poll_one() will return immediately (non-blocking) in case there is no event to process.

run_one() will block the calling thread until one event is ready to process.
```

so for example in Qt, the simplest way to integrate `asio` - is to just use the QTimer whis way:
```cpp
   ...
   io_context ioctx;
   QTimer timer;
   connect(&timer, &QTimer::timeout, [&ioctx](){ ioctx.poll_one(); });
   timer.start(1);
```
It's enough.

I want to point out that I would like to avoid using an extra thread because it's redundant.


best!

[...]

PR #476, opened 2022-09-04
Introduce Loop.

ArthurSonzogni:
It can be used to give developers a better control on the loop. Users
can use it not to take full control of the thread, and poll FTXUI from
time to time as part of an external loop.

This resolves: https://github.com/ArthurSonzogni/FTXUI/issues/474


include/ftxui/component/loop.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+namespace ftxui {
+class ComponentBase;
+
+using Component = std::shared_ptr<ComponentBase>;
+class ScreenInteractive;
+
+class Loop {
+ public:
+  Loop(ScreenInteractive* screen, Component component);
```
ArthurSonzogni, line 16: @niXman,
What do you think of this new API?

niXman: perfect!

will try it today. thanks!

niXman: @ArthurSonzogni 

this is strange, but when I exec the `custom_loop` example it only updates counters when I move mouse. is it supposed to be like that?

ArthurSonzogni: The external loop is updated 100 times per seconds. However the value is drawn to the screen only when FTXUI draw a new frame. The previous frame is invalidated when a new ftxui event is processed (e.g. a mouse move or a keyboard action)

Does this sounds wrong to you?

niXman: ah, got it!

> Does this sounds wrong to you?

no-no, please leave it as is.

thanks!

will try to use loop in that style:
```
auto screen = ScreenInteractive::FitComponent();

  // Get called periodically:
  screen.PostPeriodicTask([&] {
    // [...]
    return true; // Continue
  }, 200ms);

  auto component = ...;
  screen.Loop(component);
```

will report!

examples/component/custom_loop.cpp
```diff
@@ -0,0 +1,55 @@
+#include <stdlib.h>                   // for EXIT_SUCCESS
+#include <chrono>                     // for milliseconds
+#include <ftxui/component/event.hpp>  // for Event
+#include <ftxui/dom/elements.hpp>  // for text, separator, Element, operator|, vbox, border
+#include <memory>                  // for shared_ptr
+#include <string>                  // for operator+, to_string, allocator
+#include <thread>                  // for sleep_for
+
+#include "ftxui/component/captured_mouse.hpp"  // for ftxui
```
niXman, line 9: the Lint warns about unused header :)

include/ftxui/component/loop.hpp
niXman: ah, it's not ready yet :)

niXman:
@ArthurSonzogni thank you very much!

PR #482, opened 2022-09-25
FIX: minor mistake

LobanovaVA:
Using a static code analyzer, a suspicious place was discovered. I've fixed it.


ArthurSonzogni:
Thanks for the bug fix!
What tool did you used?

LobanovaVA:
> What tool did you used?

PVS-studio analyzer

Issue #483, opened 2022-09-27
Questions about character width.


TwinKleS-C:
Recently I'm using ftxui to make my cli app have a nice ui, but I have some doubts about the character width.
In the terminal I use (windows terminal, the font is Cascadia Code (monospace)), some characters are displayed as two latin letters wide, such as CJK characters, emoji, and this affects the layout of elements, such as the following example,
```
hbox(
  text(...), // icon text
  separator(),
  content // right content is a vbox, maybe multiline
)
```
If the text of the "icon text" element is an emoji character (which occupies twice the width in a monospaced font), the second line of content on the right element is reduced inward by one character width. like the picture below.
![Snipaste_22-09-28_01-53-54](https://user-images.githubusercontent.com/37923060/192603250-ec5d9586-7992-4284-b405-56d1c1c3c10b.png)

I don't know if this is an expected behavior or a bug?
Thanks!

ArthurSonzogni:
Here is some code pointer:
https://github.com/ArthurSonzogni/FTXUI/blob/b3aad183da2e67943d2330a396200167da523be0/src/ftxui/screen/string.cpp#L101-L109


I guess your character is not included into the set of "full width" characters. Would you mind copy-pasting it?

PR #484, opened 2022-09-27
Full width characters. Updated to unicode 13

ArthurSonzogni:
Unicode 13 contains a lot more of full characters. Add them.
    
This resolves:
https://github.com/ArthurSonzogni/FTXUI/issues/483


ArthurSonzogni:
Hi @TwinKleS-C

Could you please take a look? In particular, I would like you to verify the test expectation are correct. What character did you used for the circle?

TwinKleS-C:
> Hi @TwinKleS-C
> 
> Could you please take a look? In particular, I would like you to verify the test expectation are correct. What character did you used for the circle?

this bug seems to be fixed, thanks very much!
the character i use is ⚪🔵🟡🔴🟢🟣
Snipaste_22-09-28_10-10-50

ArthurSonzogni:
Thanks for the confirmation!

Issue #487, opened 2022-09-28
Radiobox doesn't accept enter event


jdfa:
RadioboxBase::OnEvent returns false for Event::Return: https://github.com/ArthurSonzogni/FTXUI/blob/24e04faf36bf0954272200b0bf51c4305e8b8dfc/src/ftxui/component/radiobox.cpp#L107

jdfa:
Yes, I can do that.
Do we need to pass event further if current radiobutton was selected before current event?

ArthurSonzogni:
Yes, I believe we should return true unconditionnally.

ArthurSonzogni:
Thanks for reporting this! It was fixed in #491: Radiobox now handles both Enter and Space. Closing.

PR #491, opened 2022-10-01
#487 Handled Space and Enter events for Radiobox

jdfa:
Handled Space and Enter events for Radiobox, added tests for other event handling in radiobox. 
Is there any easy way to initialize `RadioboxBase::box_ ` in testing context? It is required to test PageUp, PageDown, Tab and TabReverse events.

### DIFF
--- a/src/ftxui/component/radiobox.cpp
+++ b/src/ftxui/component/radiobox.cpp
@@ -103,8 +103,8 @@ class RadioboxBase : public ComponentBase {
 
     if (event == Event::Character(' ') || event == Event::Return) {
       *selected_ = hovered_;
-      //*selected_ = focused_entry();
       option_->on_change();
+      return true;
     }
 
     return false;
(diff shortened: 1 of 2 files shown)


ArthurSonzogni:
Thank you!

> Is there any easy way to initialize RadioboxBase::box_ in testing context? It is required to test PageUp, PageDown, Tab and TabReverse events.

No there are none. This is internal to the RadioboxBase. We should not try to test the implementation details, we should test only how the component behave, seen from the public API only.
It shouldn't be required to test keyboard only event. It is used only for the mouse.

jdfa:
> It shouldn't be required to test keyboard only event. It is used only for the mouse.

Not exactly, check this:
https://github.com/ArthurSonzogni/FTXUI/blob/c61fadd8eca94599dbcad48ded962cc0373ee966/src/ftxui/component/radiobox.cpp#L76-L81

ArthurSonzogni:
> Not exactly, check this:

You are right! Then it means the RadioBox list is bigger than the screen.
I will add some tests!

Issue #490, opened 2022-09-30
Feature request: Support text selection and word skip when input


twinstar6980:
I really like FTXUI, but I've run into some issues that are not conducive to productivity.

The first is the selection of text, in FTXUI I can't seem to select and copy text in the TUI by mouse movement, which makes me have to type based on what I see when I need text in TUI.

Secondly, there is no word skipping ability, in many terminals (I use Windows terminal, ubuntu terminal and mac terminal) can quickly move the cursor in the text via CTRL+Left/RightArrow key, which seems impossible to do in FTXUI (I did not see the relevant content in the Input component documentation), which will reduce the input efficiency when using the input component (if you need to move the cursor frequently)

I don't know if these two features will be implemented, but either way, thank you very much for developing such an awesome project❤️

ArthurSonzogni:
Both seem doable. Thanks for those suggestions!

megahallon:
I would also like to be able to select text. What kind of property is preventing this? Is it some mode in the terminal that affects everything or is it possible to allow selection of certain text?

ArthurSonzogni:
> I would also like to be able to select text. What kind of property is preventing this? Is it some mode in the terminal that affects everything or is it possible to allow selection of certain text?

I need to track mouse events, and display the Input element to "invert" the part of text selected.
There are many difficulties: UTF8 encoding, characters taking 1 or 2 cells, state management, multi-line support, etc..

This is just a matter of someone doing it. I am going to try implementing it, once I am done with [multi line support](https://github.com/ArthurSonzogni/FTXUI/blob/textarea/src/ftxui/component/text_area.cpp).

megahallon:
Before this is done, how about a mode which disables mouse capture? For my use case, mouse selection is more useful than mouse support in the UI. Is suppose this could just be a configuration flag to the interactive screen, or is it more involved?

eliasdorneles:
> The first is the selection of text, in FTXUI I can't seem to select and copy text in the TUI by mouse movement, which makes me have to type based on what I see when I need text in TUI.

Which terminal are you using?
Usually this happens because the terminal is in application mode, and the way to select in these cases is to select while pressing a modifier key. 

I'm able to select and copy by doing <kbd>Shift</kbd>+mouse selection -- should work in most cases.
On Mac, I believe it's <kbd>Option</kbd>+mouse selection.

ArthurSonzogni:
[...]

PR #494, opened 2022-10-02
Featre: Support ctrl+arrow in input.

ArthurSonzogni:
CTRL+LEFT: Move the cursor to the beginning of the word. CTRL+RIGHT: Move the cursor to the beginning of the word.

This was requested by:
https://github.com/ArthurSonzogni/FTXUI/issues/490


ArthurSonzogni:
@TwinKleS-C

Does this looks good to you?

TwinKleS-C:
Wow, this looks pretty good!
However, I found a bug where there is a possibility that an exception may occur when I use the mouse to quickly move the cursor within the input component😢
(i use msvc v19.33, screen.Loop() run in a child thread)
Snipaste_22-10-06_03-28-39

PR #497, opened 2022-10-08
Add eCAL monitor to the list of projects that are using FTXUI

brakmic-aleksandar:


ArthurSonzogni:
Really nice! Thanks for using FTXUI!

PR #498, opened 2022-10-08
Ignoring wstring_convert deprecation warnings on Clang, GCC

wflohry:


ArthurSonzogni:
It doesn't pass the test.

Maybe I can now remove those API. They have been deprecated for a while now.

ArthurSonzogni:
I adapted the patch. Does it looks good to you?

PR #499, opened 2022-10-09
fix cross compiling with MinGW

neheb:
MinGW on Linux is case sensitive. Windows is case insensitive.

Signed-off-by: Rosen Penev <rosenp@gmail.com>


ArthurSonzogni:
Thanks! Didn't know about compilers not being case insensitive. I assumed there was only a unique name for a header file.

Issue #492, opened 2022-10-01
[Windows] F1-F4 events can't be used


jdfa:
Looks like on Windows F1-F4 keys generates slightly different codes:

```
(27 79 80) F1
(27 79 81) F2
(27 79 82) F3
(27 79 83) F4
```

While predefined values are:
  ```
(27 91 79 80) F1
(27 91 79 81) F2
(27 91 79 82) F3
(27 91 79 83) F4
```

For F5-F12 it is the same, except that F11 produces ( 27 91 50 51 126 ).

ArthurSonzogni:
What terminal are you using on windows?

jdfa:
Same result in cmd, powershell and in WindowsTerminal hosting cmd and powershell.

ArthurSonzogni:
Fixed ;-)

PR #501, opened 2022-10-18
Fix F1-F4 keymapping.

ArthurSonzogni:
It was just wrong, even on Linux.

Bug:https://github.com/ArthurSonzogni/FTXUI/issues/492


ArthurSonzogni:
Hello @jdfa
Does this looks good to you?

jdfa:
Hi
Yes, looking good.
Nice that at least here it is the same for both platforms.

jdfa (approved):

PR #502, opened 2022-10-24
Fix std::ignore is in <tuple>.

5cript:
I could not build the library because std::ignore was not found.
https://en.cppreference.com/w/cpp/utility/tuple/ignore
Its in the \<tuple\> header. I added the include.


ArthurSonzogni:
Thanks!
What compiler did you use?

5cript:
clang 14 on linux.

Issue #508, opened 2022-11-14
Backspace is not working in Input example when using ssh


Saigut:
Client os: Windows 10
SSH client: xshell, Motaxterm

Remote os: ubuntu 18.04, ubuntu 20.04
Remote terminal: bash, fish-shell, zsh

Example: [Input](https://github.com/ArthurSonzogni/FTXUI/blob/master/examples/component/input.cpp)

When I run this Input example via ssh, the Backspace key is not working.

ArthurSonzogni:
Hello,
Thanks for reporting this!
This is a rather unusual setup for me you are using.

Could you try running the example: `examples/components/ftxui_example_print_key_press` ?
https://github.com/ArthurSonzogni/FTXUI/blob/master/examples/component/print_key_press.cpp

Could you please check what is the key associated sent by the terminal when typing backspace, and compare it with the webassembly version?
https://arthursonzogni.github.io/FTXUI/examples/?file=component/print_key_press

Saigut:
Hello, I just tested that. I ran `print_key_press` example and press `BackSpace` key, then:
SSH terminal output:
```
( 8 ) -> (special) 
```
WebAssembly output:
```
( 127 ) -> Event::Backspace
```

Saigut:
And in addition, when I run ssh terminal, the mouse events all have no reaction, including moving cursor, left/right click, scrolling mouse wheel, etc.

ArthurSonzogni:
Hello,
I believe I might have fixed the issue. Does it work?

Saigut:
> Hello, I believe I might have fixed the issue. Does it work?

It works! :thumbsup:

By the way, do you have any idea on why mouse events do not work on ssh terminal?

PR #510, opened 2022-11-17
Interpret 8 as 127

ArthurSonzogni:
This resolves:
https://github.com/ArthurSonzogni/FTXUI/issues/508

Apparently, this is a common issue:
https://www.cs.colostate.edu/~mcrob/toolbox/unix/keyboard.html

### DIFF
--- a/src/ftxui/component/terminal_input_parser.cpp
+++ b/src/ftxui/component/terminal_input_parser.cpp
@@ -19,6 +19,8 @@ const std::map<std::string, std::string> g_uniformize = {{
     // See https://github.com/ArthurSonzogni/FTXUI/issues/337
     // Here, we uniformize the new line character to `\n`.
     {"\r", "\n"},
+    // See: https://github.com/ArthurSonzogni/FTXUI/issues/508
+    {std::string({8}), std::string({127})},
 }};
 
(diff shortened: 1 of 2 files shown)


megahallon:
After this fix I can not build with g++ 11.2, it seems to be caused by the extra braces around the map initializer:

```
--- src/ftxui/component/terminal_input_parser.cpp.orig	2022-11-23 11:24:10.374350457 +0100
+++ src/ftxui/component/terminal_input_parser.cpp	2022-11-23 11:24:19.768526030 +0100
@@ -13,7 +13,7 @@
 namespace ftxui {
 
 // NOLINTNEXTLINE
-const std::map<std::string, std::string> g_uniformize = {{
+const std::map<std::string, std::string> g_uniformize = {
     // Microsoft's terminal uses a different new line character for the return
     // key. This also happens with linux with the `bind` command:
     // See https://github.com/ArthurSonzogni/FTXUI/issues/337
@@ -21,7 +21,7 @@
     {"\r", "\n"},
     // See: https://github.com/ArthurSonzogni/FTXUI/issues/508
     {std::string({8}), std::string({127})},
-}};
+};
 
 TerminalInputParser::TerminalInputParser(Sender<Task> out)
     : out_(std::move(out)) {}
```

PR #515, opened 2022-11-24
Fix automerge at border

megahallon:
I noticed an issue with automerging at top and left borders of the screen. Here is proposed fix with a testcase. Hope you can use this. Thanks for an awesome library.


ArthurSonzogni:
Thank you!
It looks like a perfect patch! Let's wait for the CI to run.

PR #518, opened 2022-11-27
Improve UNIX signal handling of SIGWINCH

mr-mocap:
There exists a possibility of a signal occurring during a Receiver::Receive(), while a lock is held on its mutex.  The lock is not released during the signal handler.  If we then redraw the screen inside the signal handler via the same Sender, then we try to acquire a lock on that same mutex, which effectively halts the program.

Let's move the code from inside the signal handler to outside of it.  A simple boolean flag should be fine.  We can then check for this flag in the ScreenInteractive's EventListener and perform the action there.

NOTE: The only reason this was noticed is because I changed my project's use of FTXUI from muti-threaded to a single-threaded, custom event loop.  Windows hasn't shown me this issue.

### DIFF
--- a/include/ftxui/component/screen_interactive.hpp
+++ b/include/ftxui/component/screen_interactive.hpp
@@ -66,6 +66,7 @@ class ScreenInteractive : public Screen {
 
   void HandleTask(Component component, Task& task);
   void Draw(Component component);
+  void ResetCursorPosition();
 
   void Signal(int signal);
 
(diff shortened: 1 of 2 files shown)


ArthurSonzogni:
There are several other signal handlers with the same issue. I am going to find a more generic solution.

ArthurSonzogni:
I gave it a try. Here is my partial attempt. This is not ready yet. I still see a bug when handle sigtstp

mr-mocap:
Nice improvement!

Handling signals properly is not a simple task.

ArthurSonzogni:
I observed there was some issues installing/uninstalling the screen, and some issues with nested screen. I took the opportunity to fix it here in the last couple of patches.

PR #540, opened 2022-12-28
Fix crash on empty tab container

mavam:
This PR fixes a crash when a tab container is created with empty children. All other member functions already guard for this special case, but `OnMouseEvent` did not.


ArthurSonzogni:
Thanks!

Issue #551, opened 2023-01-10
Testssuite fails to build: it can't find googletest


yurivict:
```
===>  Testing for ftxui-3.0.0.69
-- clang-tidy not found.
CMake Error at cmake/ftxui_test.cmake:18 (add_subdirectory):
  add_subdirectory given source
  "/usr/ports/devel/ftxui/work/FTXUI-3.0.0-69-g90dfcee/_deps/googletest-src"
  which is not an existing directory.
Call Stack (most recent call first):
  CMakeLists.txt:149 (include)


CMake Error at cmake/ftxui_benchmark.cmake:13 (add_subdirectory):
  add_subdirectory given source
  "/usr/ports/devel/ftxui/work/FTXUI-3.0.0-69-g90dfcee/_deps/googlebenchmark-src"
  which is not an existing directory.
Call Stack (most recent call first):
  cmake/ftxui_test.cmake:83 (include)
  CMakeLists.txt:149 (include)


-- Configuring incomplete, errors occurred!
```

googletest doesn't get fetched, and it doesn't use the pre-installed googletest either.

Version: 3.0.0-69-g90dfcee
clang-14
FreeBSD 13.1

ArthurSonzogni:
I should probably check for clang tidy and add a better error message to explain what happened.

yurivict:
No, I didn't enable FTXUI_CLANG_TIDY

ArthurSonzogni:
Okay. Then, I will take a look more closely. Thanks for reporting this!

arrowd:
@yurivict This might has something to do with the fact that our `cmake.mk` always passes `-DFETCHCONTENT_FULLY_DISCONNECTED:BOOL=ON` in `CMAKE_ARGS`.

yurivict:
It should be possible to build tests with pre-installed googletest.

arrowd:
Yes, sure, I'm just guessing why @ArthurSonzogni  may not seen this before.

yurivict:
I see.

yurivict:
> gtest from packages does not come with GTestConfig.cmake to find the library.

Yes it does:
```
$ pkg info -l googletest | grep GTestConfig
	/usr/local/lib/cmake/GTest/GTestConfig.cmake
	/usr/local/lib/cmake/GTest/GTestConfigVersion.cmake
```

It can be made optional to use externally installed GTest.

ArthurSonzogni:
[...]

PR #552, opened 2023-01-11
cmake: support gtest from the package manager

ArthurSonzogni:
Some developers would be happier with the gtest version provided from their package manager. Use it if it is installed the package provide cmake support.

Fixed: https://github.com/ArthurSonzogni/FTXUI/issues/551


ArthurSonzogni:
@yurivict / @arrowd
Could you please confirm [this](https://github.com/ArthurSonzogni/FTXUI/pull/552) fixes: https://github.com/ArthurSonzogni/FTXUI/issues/551

(I can't test it locally myself, because the ubuntu package manager do not provide the proper cmake finder)

yurivict:
I have benchmark-1.7.1 installed.

ArthurSonzogni:
Thanks!
I did the same for google-benchmark. Does it work now?

PR #553, opened 2023-01-11
Readme.md: change vcpk.info to vcpkgx.com

alexm622:
The vcpkg.info site is now defunct.
[vcpkgx.com](https://github.com/vcpkgx/vcpkgx.github.io) provides a similar if not identical service to what vcpkg.info used to provide.


ArthurSonzogni:
Thanks!

Issue #569, opened 2023-02-06
[Feature Request] Add hook to execute custom code in ScreenInteractive::Loop


Mestkon:
I have an application where I would like to drive the event system of another framework in the loop. Is it possible to add a simple overload to the Loop function where I can give it a lambda to be executed each loop iteration? Or another function to give it a lambda?

Example:

    ScreenInteractive screen;
    Component cmp;
    screen.Loop(cmp, [&]() { other_framework.do_update(); });

Or:

    ScreenInteractive screen;
    Component cmp;
    screen.OnUpdate([&]() { update_stuff(); });
    screen.Loop(cmp);

ArthurSonzogni:
Hello,

Here is a solution:
```cpp
  Loop loop(&screen, component);

  while (!loop.HasQuitted()) {
    other_framework.do_update();
    loop.RunOnce();
    std::this_thread::sleep_for(std::chrono::milliseconds(10));
  }
```

If you want FTXUI to draw a new frame, you have to call `RequestAnimationFrame` or post a custom event.
```cpp
  if (other_frame_worker.do_update()) {
    screen->RequestAnimationFrame(); // 1st option
    screen->PostEvent(Event::Custom); // 2nd option
  }
```

--------------

alternatively with a thread:
```cpp
  auto screen = ScreenInteractive::Fullscreen();

  // Execute a task periodically.
  std::atomic<bool> submit_periodic_task = true;
  std::thread submit_periodic_task_loop([&] {
    while (submit_periodic_task) {
      using namespace std::chrono_literals;
      std::this_thread::sleep_for(0.5s);
      screen.PostTask([&]  {
        // Executed on the screen's thread (the main thread)
        other_framework.do_update();
      });
  });

  screen.Loop(main_renderer);

  submit_periodic_task = false;
  submit_preriodic_task_loop.join();
```

Does something solve your problem?

Mestkon:
It looks like the above might solve the problem yes. Thank you. 

I made a pull request to simplify this problem in the future if my simplified solution fits the design goals of the library

ArthurSonzogni:
`Loop::RunOnce()` (see the snippet above) lets you drive your own framework between frames, so I'll close this. Thanks for the idea!

PR #570, opened 2023-02-06
Feature: Add closure to be executed each loop iteration to ScreenInteractive::Loop

Mestkon:
This resolves:
#569


ArthurSonzogni:
Thank you! This is greatly appreciated!
I don't think we want to propose such API, given developers have alternatives.

Long term, I would like to make the task scheduler a little bit more useful and propose delayed task. I guess the current patch correspond to some `PostIdleTask`.

Issue #580, opened 2023-02-14
Feature: ResizableSplit with custom separator


pllee4:
I am enjoying using `ResizableSpilt` for layout. Feels like in some cases, it would be good to have custom separator, for example, if I would like to have the separator invisible, I can simply pass element of `separatorEmpty()`.

Thinking perhaps it would be good if we can have one more option for the interface.

mavam:
Another use case: colorize the separator in a lighter grey tone so that it doesn't show up so much.

ArthurSonzogni:
Thanks! This is grealy appreciated!
I like it. I wonder if there are other ways to make it customizable easily.

I think I will take your patch, but modify the interface to allow more.

mavam:
Great to see this merged. Thanks a lot! 🙏

PR #581, opened 2023-02-14
Feature: ResizableSplit with custom separator

pllee4:
This resolves:
    https://github.com/ArthurSonzogni/FTXUI/issues/580

PR #583, opened 2023-02-15
Feature resizable spilt with custom separator

ArthurSonzogni:
Create from: https://github.com/ArthurSonzogni/FTXUI/pull/581

This resolves:
https://github.com/ArthurSonzogni/FTXUI/issues/580

### DIFF
--- a/src/ftxui/component/resizable_split.cpp
+++ b/src/ftxui/component/resizable_split.cpp
@@ -14,10 +14,14 @@ namespace {
 
 class ResizableSplitLeftBase : public ComponentBase {
  public:
-  ResizableSplitLeftBase(Component main, Component child, int* main_size)
+  ResizableSplitLeftBase(Component main,
+                         Component child,
+                         int* main_size,
+                         Element separator)
       : main_(std::move(main)),
         child_(std::move(child)),
-        main_size_(main_size) {
+        main_size_(main_size),
+        separator_(separator) {
     Add(Container::Horizontal({
         main_,
         child_,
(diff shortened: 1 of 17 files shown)


ArthurSonzogni:
- I deduplicated the ResizableSplit class.
- I introduced ResizableSplitOptions, so that folks can use designated initializer to tweak the option they want.
- I stopped passing an Element, and replaced it by a function<Element()>. The reason is that an `Element` isn't meant to be used drawn on screen more than once.

I think it still requires some work. I would be very interested plumbing the mouse distance / position relative to the split to make some particular effects.

src/ftxui/component/resizable_split.cpp
```diff
@@ -264,20 +148,24 @@ class ResizableSplitBottomBase : public ComponentBase {
 /// int left_size = 10;
 /// auto left = Renderer([] { return text("Left") | center;});
 /// auto right = Renderer([] { return text("right") | center;});
-/// auto split = ResizableSplitLeft(left, right, &left_size);
+/// auto split = ResizableSplitLeft(left, right, &left_size, separatorDouble());
```
pllee4, line 151: minor: I guess can revert those changes on interface which was "polluted" by me for `ResizableSplit...` for the example code in the documentation.

pllee4:
Looks great for the changes, thanks!

ArthurSonzogni:
Sorry for the delay. I wanted to add a way to visualize the mouse was close to the resizeableSplit, by passing kind of coordinates to the function. I think I will find another way in the future, or maybe break the current API.

PR #592, opened 2023-03-08
Feature: `colorgrad` and `bgcolorgrad` decorator.

v1nns:
Based on the existing color decorators, create new ones to apply a gradient effect on the DOM.


ArthurSonzogni:
Thank you!
The code looks clean at first glance. I will take a look later tonight.

I wonder if we should handle:
- The gradient direction,
- Multiple colors along the axis,
- Gradient along the other direction like a 4 corners gradient.

Maybe that would be overkill(?)

```cpp
  component->Render() |
      bgcolorGradient(LinearGradient(90, {
                                             {Color::Red, 0},
                                             {Color::Green, 0.5},
                                             {Color::Blue, 1},
                                         }));

component->Render() | bgcolorGradient(45, LinearGradient({Color::Red, Color::Blue}));
```

v1nns:
I think I've got your idea, that would make it really similar to CSS linear-gradient effect.

In my opinion, the first two are nice to have it. And the last one is a little overkill, but it would be a visually appealing effect to have it! 

I will check later why the new tests have failed with Windows MSVC. :sweat:

v1nns:
Hi, based on your previous comment, I did some changes on the code. Please, take a look and tell me what you think.

As I didn't see any header files (other than **elements.hpp**) for other decorators, I don't know what would be the right approach to create the LinearGradient struct.

And also, I put an assert inside the hidden GradientColor class constructor, but again, it may exist a better approach but I would like to ask your opinion about this.

ArthurSonzogni:
Awesome!

Sorry for the lack of interaction from my side recently. I was quite busy.
This is looking good to me. I will take over and tweak it a bit.

v1nns:
Thanks! 

That's fine, don't worry about that.
Alright, I'm looking forward for this feature (I intend to use it on a TUI music player that I've been developing :heart_eyes:).

And thank you for your attention.

ArthurSonzogni:
Hi @v1nns,

I think I am done with my tweaks. It reached a state I like.

I added:
- Fixed Color::Interpolate. We are now using gamma correction to get "correct" colors.
- Support for abritrary angle in degree.
- Support for stops with optional position. When not provided, the position is interpolated from nearby stops.
- Added two examples.
- Add a section to the documentation.
- Moved the tests into a dedicated file. Added a few checking about the API.
- - Support degenerated user provided input:
  - Non sorted positions.
  - Position outside of |0,1]
  - Gradient with less than 2 stops.

If you will, I will let you review the final patch. Could you please take a look?

ArthurSonzogni:
(I will land it now, because I would like to show it using the https://arthursonzogni.github.io/FTXUI/examples/, but I am still interested getting feedback if you have)

v1nns:
Hi @ArthurSonzogni,

Sorry about the delay.

The idea of adding stops and to use the gamma correction were nice.
I gotta be honest with you, the final result is simply amazing, you did a great job!

About the review itself, I don't have much to say other than minor typos (I'll add comments in a review, but it's not a big deal at all).

Thank you for your work!

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ current (development)
 ### Dom
 - Feature: Add the dashed style for border and separator.
 - Feature: Add colored borders.
+- Feature: Customize with gradient color effect. Add the following decorators:
+  - `colorgrad`
+  - `bgcolorgrad`
```
v1nns, line 16: color(LinearGradient) and bgcolor(LinearGradient)

PR #600, opened 2023-03-23
Fixed all signed/unsigned warnings in lib and fixed some (critical) mistakes

LostInCompilation:
I've gone through the code and have fixed all (hundreds, lol) signed/unsigned warning and some critical mistakes. There are no breaking changes, every warning is handled with `static_cast`, casting to a safe type which is expected. Also there are no changes to the logic of the code or the behaviour.
Especially the signed/unsigned usage was complete and utter randomness. Most of the warnings were caused by random casts, random changes in signed/unsigned (one line uses `int`, two lines below `unsigned int` got used for the exact same function call and arguments for no reason) or just casts from signed to unsigned and back to signed within 2 lines for no reason.
Also I fixed some mistakes which could potentially be critical in some situations. See the changelog below.

**Please note:** There is only one single change of a function signature:
```
elements.hpp Line 81: Element spinner(size_t charset_index, size_t image_index);
spinner.cpp Line 280: Element spinner(size_t charset_index, size_t image_index)
```
I changed the first paramter (charset_index) from `int` to `size_t`. This does not break anything, since the function internally only uses size_t. Also Apps using FTXUI are compatible with this change.

This was a bit of work, I hope this can cleanup the lib a little bit so the `NOLINT` hack is no longer necessary. I tested the lib, built the tests and also checked all examples.

If you have suggestions or questions, feel free to comment.

### Full Changelog
- Fixed many, many unsigned/signed conversion warnings
- Fixed unsafe float comparison by epsilon comparison
- Fixed wrong semicolons at the end of function blocks (void foo() {...}; <-- this one)
- Fixed wrong parameter documentation name ("(AT)params" instead of "(AT)param")
- Fixed double to float conversion warnings
- Changed old C-style casts to C++ static_cast
- Fixed a hard (intention) bug of misused comma instead of semicolon (dropdown.cpp Line 31). While the old version should work, the comma seems to not be intended


ArthurSonzogni:
Thank you for this hard work! That's really appreciated.
I will be able to take a closer look tomorrow.

LostInCompilation:
I think I uploaded the wrong string files. I will check the failing tests tomorrow, sorry. I'll update you.

ArthurSonzogni:
This is introducing some bugs, I think I am going to split this PR into a few smaller ones, because there are parts that should clearly be merged right now. They don't need to be tied to this PR.

LostInCompilation:
Thank you for your work. Do you mean bugs in string.cpp? Or somewhere else?
Let me know how I should proceed with this PR. I did not have the time yet to fix the issues with string.cpp, still on it.

LostInCompilation:
@ArthurSonzogni I've checked the new codebase. There seems to be some cast/signed/unsigned issues again. Some casts I added in this PR are not merged to master. Should I go over them again?
Except for strings.cpp/hpp since casting there introduces some major bugs, I first need to investigate that.

LostInCompilation:
I think I will make a PR with several commits. One commit for each file I fixed signed/unsigned stuff again. That should be easier to merge for you if one commit is not right.
I'm not really sure why you left out some safe castings (like index variables on arrays where "i" in arr[i] is a signed type which I casted to unsigned). Are there some bugs I missed?

ArthurSonzogni:
Thanks @LostInCompilation,
I kept the ones that were fixing potential bugs. However I didn't for the other where the static_cast<T> syntax was hurting readability.

Why are you interested in converting them? Is there any compiler option you would be interested adding/removing from:
https://github.com/ArthurSonzogni/FTXUI/blob/master/cmake/ftxui_set_options.cmake#L58-L77

?

LostInCompilation:
Oh yeah, static_cast inside an array index access is hurting readability a lot. I mainly used just static_cast since fixing unsigned/signed warning would introduce many changes in the codebase. So casting was just a "hack".

It's not mandatory to have signed/unsigned warnings fixed. I just thought it's cleaner, since you could enable more warnings. I'm mainly interested in solving all `-Wsign-conversion` issues.
But if you want to leave it how it is now, thats perfectly fine, let me know. If not I can make a PR where I fix some of the errors which don't hurt readability a lot.

PR #603, opened 2023-03-24
Feature: Added nix build support

### LINKED ISSUE #598: [Feature requests] Nix build system support
Would you be interested into adding nix build support for this project? I would love to use this and I would prefer to add it as nixpkg dependency.

I can try create some flake.nix, but I would love to know if it will get merged or if I will be just wasting my time before hand.

https://github.com/NixOS/nixpkgs
https://github.com/NixOS/nix
https://repology.org/

Thanks for reply.

dev-null-undefined:
Not sure if this will work, but basic example seems to be working.
Can you confirm that this lib has no dependencies and no absolute path.

Closes #598

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,43 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1678901627,
+        "narHash": "sha256-U02riOqrKKzwjsxc/400XnElV+UtPUQWpANPlyazjH0=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "93a2b84fc4b70d9e089d029deacc3583435c2ed6",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1679734080,
+        "narHash": "sha256-z846xfGLlon6t9lqUzlNtBOmsgQLQIZvR6Lt2dImk1M=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "dbf5322e93bcc6cfc52268367a8ad21c09d76fea",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixpkgs-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "flake-utils": "flake-utils",
+        "nixpkgs": "nixpkgs"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,61 @@
+{
+  description = "C++ Functional Terminal User Interface library";
+
+  inputs = {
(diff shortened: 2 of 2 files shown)


ArthurSonzogni:
Thanks!

---

+CC @IvarWithoutBones who made the default.nix in nixpkgs:
https://github.com/NixOS/nixpkgs/pull/181000
If possible, could you tell me if this patch makes sense? If you have some suggestions, they would be very appreciated. I don't know much about nix yet.

---

@dev-null-undefined, thank you very much!

I don't know nix. For now, I did:
- install nix
- enable flakes experimental feature
- tried `nix build .` with errors about not defaultPackage defined. I guess I am following the wrong path.

To merge this patch, and do some due dilligence. I would like to verify myself. What step should I follow to depend on ftxui using this config and nix?

dev-null-undefined:
I decided against defining this as default package for you to build this you would have to run `nix build .#ftxui`

dev-null-undefined:
But it could be that I am mistaking int not defining defualtPackage it would probably also make sense to add default.nix for those whiteout nix experimental.

dev-null-undefined:
It would also make sense to a change the default cmakelist (in the docs) since when building under nix you do not have network connection for purity reasons.

dev-null-undefined:
> Thanks!
> 
> ---
> 
> +CC @IvarWithoutBones who made the default.nix in nixpkgs:
> https://github.com/NixOS/nixpkgs/pull/181000
> If possible, could you tell me if this patch makes sense? If you have some suggestions, they would be very appreciated. I don't know much about nix yet.
> 
> ---
> 
> @dev-null-undefined, thank you very much!
> 
> I don't know nix. For now, I did:
> - install nix
> - enable flakes experimental feature
> - tried `nix build .` with errors about not defaultPackage defined. I guess I am following the wrong path.
> 
> To merge this patch, and do some due dilligence. I would like to verify myself. What step should I follow to depend on ftxui using this config and nix?

Why didn't you tell me about that nix config in nixpkgs :D i didn't even notice it until now :joy:

ArthurSonzogni:
> I decided against defining this as default package for you to build this you would have to run nix build .#ftxui

It works! Thanks!

---

> Why didn't you tell me about that nix config in nixpkgs :D i didn't even notice it until now joy

I discovered it after your patch, while trying to learn nix.

---

Question: Why do we need a devshell?

dev-null-undefined:
> Question: Why do we need a devshell?

For testing without having to compile the ftxui your self (nix will use the cached version instead), you can also append all build tools that are used by this for example doxygen cmake gcc and others if needed (graphviz, valgrind, formatter?) and make it into very simple single command dev environment.

And also because of the way that nix builds packages, it will not reuse object files from past compilations which will make it compile the whole project from scratch meaning that even the simplest change can take several minutes to build.

And it also provided a way to specify custom commands such as the compile (alias, function) that I have provided as example.

This is just a draft for that exact reason, this is how I would do it but I do not know this project very well for me to make any of those decisions.

dev-null-undefined:
But dev shells are not the main objective over here, those could be added only to the example project if you feel like they are more distraction then they are benefit to this project.

Mainly I would love the CMakelist.txt to change in the docs and maybe even the flake.nix, even if it would be without the dev shells, being merged so that we can get the new version on nix as well and others can easily use this lib with nix.

ArthurSonzogni:
I am now sufficiently up-to-date with nix to understand the current patch. What do you think?

PR #608, opened 2023-03-27
Casting and documentation fixes

LostInCompilation:
### **EDIT: PR is complete.**

---

Just a follow up for #600 as a preview of what I would change again. You can check the changes and let me know if something like this is OK for you or you want to leave it like it is now.
~~I'm still working on this PR, please leave it open for now. I will notify you when done.~~

If you don't want to merge these changes just let me know 👍🏼
I will make a separate commit for every file with a small explanation of the changes. I think that's easier for you to pick some commits if you only want to merge a few.

One extremely useful warning to add would be `-Wdocumentation` since sometime there are a few documentation issues. However I could not find this flag for gcc, only clang (which I'm using on macOS). Should clang be added as an extra else case here? https://github.com/ArthurSonzogni/FTXUI/blob/master/cmake/ftxui_set_options.cmake#L58-L77


LostInCompilation:
@ArthurSonzogni This PR is now done. I only fixed documentation issues and old C-Style casts to C++ static_cast. I left the signed/unsigned warning as they are.
However could you please double check, if it is intended that `void ComponentBase::SetActiveChild(ComponentBase* child)` in `component.cpp` is actually an empty function?
I made a small hack to prevent the unused parameter error. In the old state, clang complains about wrong documentation, since it cannot find the parameter `child`.

LostInCompilation:
Is there an issue with using 'static_cast'? Or did you remove it for readability?

src/ftxui/component/component.cpp
```diff
@@ -166,7 +166,7 @@ bool ComponentBase::Focused() const {
 /// @brief Make the |child| to be the "active" one.
 /// @param child the child to become active.
 /// @ingroup component
-void ComponentBase::SetActiveChild(ComponentBase* /*child*/) {}
+void ComponentBase::SetActiveChild([[maybe_unused]] ComponentBase* child) {}
```
LostInCompilation, line 169: This is absolutely the better way to solve the unused parameter error. However I decided against it, since it requires C++17. I'm not sure if FTXUI requires C++17 anyway, but if not this could be a breaking change for projects which need to "migrate" to C++17.

ArthurSonzogni: FTXUI requires C++17 privately and publicly.. So we are fine.

ArthurSonzogni:
Thanks for the `-Wdocumentation` that pretty neat! I added it as ftxui private build flags.

Yes, sorry, I removed most of the new static_cast<>.
I tried adding -Wconversion and -Wsign-conversion flags to see what I would get. I was not very happy with the results. It degrades readability to much..

ArthurSonzogni (approved):

LostInCompilation:
>Yes, sorry, I removed most of the new static_cast<>.
I tried adding -Wconversion and -Wsign-conversion flags to see what I would get. I was not very happy with the results. It degrades readability to much..

I think you misunderstood me. I only replaced old C style casts with static_cast to get rid of the old casts. I did not touch any signed/unsigned stuff, since you removed almost everything about that in my last PR already. The old C style casts should just not be used anymore, they became obsolete 13 years ago. The current approach by you using a constructor syntax `int(42L)` seems to be even worse since it adds so much room for misunderstandings and performance issues. It's your lib and you can freely do what you want with it. But omitting standard C++ because readability is a bit lower shouldn't be the way to go in my opinion. Also static_cadt becomes perfectly fine to read once you got used to it, and generally got used to huge templated names from the C++ std lib.

ArthurSonzogni:
@LostInCompilation

About the difference in between [c-style, functional-style, and static_cast](https://en.cppreference.com/w/cpp/language/explicit_cast). They are all equivalent when dealing with non-const numbers. They should produce the exact same binary output There are no performance issues. They are not obsolete. They are only not the preferred way.

If we want enforce not using them, we should let the compiler enforce the behavior, via build flag like: `-Wold-style-cast`. Same thing for `-Wsign-compare`, `-Wconversion`. I tried once, to apply them, and I was pretty sad about the result. It took me ~4 hours refactoring code, but I prefered dropping what I did. It hurt readability too much. Maybe later.

If we want to get it at the end, I think I would be happier updating FTXUI API to use size_t more often, so that type matches, and so that we don't have to make the conversion. On the other side, I am following the [Google C++ coding style](https://google.github.io/styleguide/cppguide.html#Integer_Types), and expose `int` in the API.

PR #611, opened 2023-04-03
Add link to openSUSE package

jubalh:


ArthurSonzogni:
Thanks for working on this!
I took a look at the package definition. This looked great as far as I can tell.

jubalh:
>  I took a look at the package definition. This looked great as far as I can tell.

Thanks! I would have one question: the libftxui-component, -dom and -screen will always be bumped at the same time, yes?

ArthurSonzogni:
> Thanks! I would have one question: the libftxui-component, -dom and -screen will always be bumped at the same time, yes?

What do you mean by "bumped at the same time"?   (I am not an English-native, and don't see any clear translation)

-----


By the way, cmake/cpack are already generating the rpm package:
https://github.com/ArthurSonzogni/FTXUI/releases/download/v4.0.0/ftxui-4.0.0-Linux.rpm
I was hoping defining it would help some folks. Not sure if it helped here, or could help here.

jubalh:
> What do you mean by "bumped at the same time"? (I am not an English-native, and don't see any clear translation)

Sorry for being vague. I meant whether all those versioned shared objects will be increased together. Like it wont happen that there can be liftxui-component-5.0.0.so and libftxui-dom-4.0.0 in the same release.

> I was hoping defining it would help some folks. Not sure if it helped here, or could help here.

It might help some people on RPM based systems. But distributions usually want to build those RPMs themselves. And writing the spec file was easy enough for ftxui, which is a good sign for your library :)

PR #620, opened 2023-04-17
Use globally set CMAKE_CXX_STANDARD, if it is set

lefticus:
This makes FTXUI a better citizen for things like CPM, so that you can have consistent c++ standards settings across your dependencies.

### DIFF
--- a/cmake/ftxui_set_options.cmake
+++ b/cmake/ftxui_set_options.cmake
@@ -43,9 +43,17 @@ function(ftxui_set_options library)
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
   )
 
+  # Play nicely if we are being consumed by another project
+  # and use their CMAKE_CXX_STANDARD. Otherwise, fall back to 17
   # C++17 is used. We require fold expression at least.
+  if (DEFINED CMAKE_CXX_STANDARD)
+    set(CXX_STANDARD ${CMAKE_CXX_STANDARD})
+  else()
+    set(CXX_STANDARD 17)
+  endif()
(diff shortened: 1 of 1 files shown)


ArthurSonzogni:
Thanks!

What are the benefits of having a consistent c++ standard across all the dependencies?

The cons: some users won't be able to depend on FTXUI if they specify a standard lower than c++17

lefticus:
Compiling with different standards can lead to ODR violations. I discovered
a particular issue while compiling my code with 20 and ftxui with 17 while
upgrading my Best Practices template repository.

Someone could still choose to risk a 14/17 combination with ftxui, they
would just need to set the CXX_STANDARD immediately before consuming your
CMakeLists, and restore it immediately after (I have done this in other
projects by making use of sub folders)

I have confirmed that other projects (like spdlog) behave similarly to what
I've proposed for yours.

We could add a check to make sure the standard is at least 17, and bump it
to 17 if it is not? But I must say that some of these subtle issues around
things like ODR violations are concerning me more and more as we talk about
the security of C++.

ArthurSonzogni:
Thanks for the explanation!

We might want to tweak it a bit in the future, but the current patches sounds good to me. Thanks for making it!

SpaceIm:
The proper fix should be to add `target_compile_options(${library} PUBLIC cxx_std_17)` and remove all references to `CMAKE_CXX_STANDARD`.
https://github.com/ArthurSonzogni/FTXUI/pull/156 got it right, but https://github.com/ArthurSonzogni/FTXUI/pull/552 broke it and afterwards it was a succession of incorrect workarounds.

Issue #626, opened 2023-04-24
Arrows not working - Event detects reset, not set


JosefLitos:
Arrow keys (without other modifiers) are denoted with `O` at the beginning of escape sequence, but the `O` is missing from these lines:
https://github.com/ArthurSonzogni/FTXUI/blob/a60201317916f20da676cbd807169b9e543b2275/src/ftxui/component/event.cpp#L53-L56

(should be `\x1BOA`,`\x1BOB`,`\x1BOC`,`\x1BOD`)

Also (this is a nit pick, so please feel free to ignore) `Esc` keycode is usually written in the form `\033`, which is the same length and thanks to its number-only representation is slightly easier to distinguish from the rest of the escape sequence (like here, compare `\x1BOB` vs `\033OB`), this also avoids errors when for example `A` would begin the escape sequence as it would get compiled as part of the hex code.

ArthurSonzogni:
Hello!
What terminal are you using?

ArthurSonzogni:
I see! I guess your terminal or one of your app is setting it differently.

We can easily fix this. There are several ways.

PR #627, opened 2023-04-25
Support arrow keys in application mode.

ArthurSonzogni:
Depending on the Cursor Key Mode (DECCKM), the terminal sends different escape sequences:

Key     Normal    Application
-----   --------  -----------
Up      ESC [ A   ESC O A
Down    ESC [ B   ESC O B
Right   ESC [ C   ESC O C
Left    ESC [ D   ESC O D
Home    ESC [ H   ESC O H
End     ESC [ F   ESC O F

I decided not to update the current mode to fit what FTXUI parse, but instead support parsing both. We convert the sequences from the "application mode" into the "normal mode".

Bug:https://github.com/ArthurSonzogni/FTXUI/issues/626 Fixed:https://github.com/ArthurSonzogni/FTXUI/issues/626

### DIFF
--- a/src/ftxui/component/terminal_input_parser.cpp
+++ b/src/ftxui/component/terminal_input_parser.cpp
@@ -19,8 +19,31 @@ const std::map<std::string, std::string> g_uniformize = {
     // See https://github.com/ArthurSonzogni/FTXUI/issues/337
     // Here, we uniformize the new line character to `\n`.
     {"\r", "\n"},
+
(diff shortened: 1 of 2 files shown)


ArthurSonzogni:
Hi @JosefLitos,
Could you please take a look?

JosefLitos:
Works great! :+1:

ArthurSonzogni:
Thanks!

Issue #637, opened 2023-05-15
Stall on exit due to parentless Component


nwfistere:
I'm working on a chat server with a client that utilizes ftxui as the interface. I created several custom components that inherit from `BaseComponent`. I've ran into an issue that on shutdown, the client hangs. I've found it's due to one of my custom components having a TabContainer as a child, but the TabContainer does not have a parent. Causing an infinite loop in `DetachAllChildren()`.

Using gcc 12.2.1 on ubuntu

I made the following changes in component.cpp and this is the output I'm getting.

```diff
@@ -61,23 +61,29 @@ void ComponentBase::Add(Component child) {
 /// @see Parent
 /// @ingroup component
 void ComponentBase::Detach() {
+  std::cout << "Detach(): detaching " << this << "\n";
   if (parent_ == nullptr) {
+    std::cout << "Detach(): parent is nullptr " << this << "\n";
     return;
   }
   auto it = std::find_if(std::begin(parent_->children_),  //
                          std::end(parent_->children_),    //
                          [this](const Component& that) {  //
-                           return this == that.get();
+                            std::cout << "Detach(find_if)(): returning " << this << " == " << that.get() << " " << (this == that.get() ? "true" : "false") << "\n";
+                            return this == that.get();
                          });
   ComponentBase* parent = parent_;
   parent_ = nullptr;
+  std::cout << "Detach(): erasing " << *it << "\n";
   parent->children_.erase(it);  // Might delete |this|.
 }
 
 /// @brief Remove all children.
 /// @ingroup component
 void ComponentBase::DetachAllChildren() {
+  std::cout << "DetachAllChildren(): current object: " << this << "\n";
   while (!children_.empty()) {
+    std::cout << "DetachAllChildren(): detaching " << children_[0].get() << "\n";
     children_[0]->Detach();
   }
 }
```
**Output:**
My custom component is the parent (0x7a82b0) // Note the b vs 6 for the address, they are different.
TabContainer is 0x7a8260 
```bash
DetachAllChildren(): current object: 0x7a82b0
DetachAllChildren(): detaching 0x7a8260
Detach(): detaching 0x7a8260
Detach(): parent is nullptr 0x7a8260
DetachAllChildren(): detaching 0x7a8260
Detach(): detaching 0x7a8260
Detach(): parent is nullptr 0x7a8260
DetachAllChildren(): detaching 0x7a8260
Detach(): detaching 0x7a8260
[...]

PR #638, opened 2023-05-16
Delete copy constructor and copy operator

nwfistere:
Delete copy constructor and copy operator to prevent odd behavior during object deletion. #637


ArthurSonzogni:
Excellent! Thanks!

PR #642, opened 2023-05-17
Add pkg-config file.

hzeller:
This file describes the include and lib location and makes it possible to use the library in build systems other than CMake, e.g. in basic Makefiles.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 3.11)
 project(ftxui
   LANGUAGES CXX
   VERSION 4.0.0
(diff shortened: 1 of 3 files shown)


ArthurSonzogni:
Excellent! Thank you very much!

hzeller:
Would this warrant an intermediate release ?

If this would be e.g. `4.0.1` or something, then packagers could pick that up and provide their distributions with `pkg-config` files.

(I am currently updating the nix package, and it would be excellent to have the pkg-config support)

ArthurSonzogni:
> Would this warrant an intermediate release ?


@hzeller: I released created today `v4.1.0`: https://github.com/ArthurSonzogni/FTXUI/releases/tag/v4.1.0

PR #643, opened 2023-05-19
build: Don't overwrite cmake version config

BurningEnlightenment:
The current `install(EXPORT` command overwrites the `ftxui-config-version.cmake`, i.e. no version compatibility checks happen.

The idiomatic name for the export config is `<package-name>-targets.cmake` or `<PackageName>Targets.cmake` (as shown in the [cmake docs](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-a-package-configuration-file)).


BurningEnlightenment (commented):
Generally you should enquote paths with variables as CMake splits them into two/many arguments if the variable contains spaces

PR #645, opened 2023-05-19
Fix MSVC possible loss of data warning

robinlinden:
warning C4244: '=': conversion from '_Ty' to 'float', possible loss
    of data

Looks like the rest of the file already writes "0" like this when assigning to LinearGradient types, so probably just an oversight.


ArthurSonzogni:
Thanks!

PR #647, opened 2023-05-20
CMake: use compile_features to set min C++ standard

SpaceIm:
As pointed out in https://github.com/ArthurSonzogni/FTXUI/pull/620#issuecomment-1555930977, compile_features is the proper way to set min C++ standard in CML of a project, specially for libraries.
Moreover compile features are propagated to CMake config file, unlike CXX_STANDARD property.

If needed, it can be overridden externally by users with CMAKE_CXX_STANDARD, which has precedence over compile feature.

I don't know exactly what is the minimum C++ standard of FTXUI libraries, tests, examples etc. There was CMAKE_CXX_STANDARD 20 for tests and fuzzers, and 17 for libraries and examples. I've assumed C++17 was the minimum standard of all targets, and I've set cxx_std_17 to all libraries and executables.

### DIFF
--- a/cmake/ftxui_fuzzer.cmake
+++ b/cmake/ftxui_fuzzer.cmake
@@ -14,10 +14,7 @@ function(fuzz source)
   target_link_libraries(${name} PRIVATE component)
   target_compile_options(${name} PRIVATE -fsanitize=fuzzer,address)
   target_link_libraries(${name} PRIVATE -fsanitize=fuzzer,address)
-  set_target_properties(${name} PROPERTIES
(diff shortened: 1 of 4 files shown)


ArthurSonzogni:
Thanks!
If 'CMAKE_CXX_STANDARD' takes precedence as you said, then this might work.

@lefticus what do you think about it?

SpaceIm:
So indeed, tests require C++20. I think C++17 is sufficient for fuzzers.

ArthurSonzogni:
Thanks!

PR #654, opened 2023-05-29
build: Fix various CMake bugs/nits

BurningEnlightenment:
CMake variable expansions `${...}` _must_ be quoted if the contents might contain a ` ` (space) character. Otherwise CMake will split the content into multiple arguments. This is especially important for filepaths.

I also fixed the pkg-config install directory. See the commit messages for further details.


ArthurSonzogni (approved):
Thanks!

The commits description make them easy to understand. They all look good to me.

BurningEnlightenment:
@ArthurSonzogni these fixes should probably be applied to `support-v4`, too.

ArthurSonzogni:
>  @ArthurSonzogni these fixes should probably be applied to support-v4, too.

Added!

I just released v4.1.1 twelve hours ago, so I won't do another minor release immediately, but at least it will be included in the next one.

PR #656, opened 2023-05-29
Make a table of projects with stars

ei-grad:


ei-grad:
Doesn't look good actually. Though it was useful to look on the list of repo stars. Generated it with chatgpt :).

PR #659, opened 2023-05-30
Optimize inserts in vector and refactor const reference objects

GermanAizek:
@ArthurSonzogni,
hi, I think this optimization will be very useful, and I did a minor refactoring.


ArthurSonzogni:
Thanks! That's appreciated.

I don't believe any of this can affect benchmarks. This is mostly premature optimizations. These vectors are not used in hot paths, they are usually small, and contains small sized elements, usually pointers.

Anyway, it doesn't really hurt adding some `std::vector::reserve()` here and here. The only drawback is code readability, indeed its adds noise around the important lines. So this is mostly a tradeoff.  So, I decided to merge this patch after removing the parts I liked less (see patchset 2)

Thank you!

PR #660, opened 2023-06-01
fix some warnings

ProkopRandacek:
They were spamming my build log :D

### DIFF
--- a/examples/dom/color_info_sorted_2d.ipp
+++ b/examples/dom/color_info_sorted_2d.ipp
@@ -15,10 +15,9 @@ std::vector<std::vector<ftxui::ColorInfo>> ColorInfoSorted2D() {
   }
 
   // Sort info_color by hue.
-  std::sort(info_color.begin(), info_color.end(),
-            [](const ftxui::ColorInfo& A, const ftxui::ColorInfo& B) {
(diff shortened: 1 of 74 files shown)


ArthurSonzogni:
Thanks!

If you will, could you please indicate how to add these errors to the CI? Are you using a particular compiler?

On the CI, they are behind the `FTXUI_DEV_WARNINGS` option.

examples/dom/color_info_sorted_2d.ipp
```diff
@@ -33,22 +32,22 @@ std::vector<std::vector<ftxui::ColorInfo>> ColorInfoSorted2D() {
...
           best_index = j;
         }
       }
-      std::swap(column[i], column[best_index]);
+      std::swap(column[i + 1], column[best_index]);
     }
   }
 
-  return info_columns;
+  return std::move(info_columns);
```
ProkopRandacek: writing std move here disables copy elision and forces a call to the move constructor instead (as far as i know)

ArthurSonzogni: You are right ;-)

ProkopRandacek: then why did you revert it?

PR #670, opened 2023-06-06
Separate focused entry option from MenuOptions class, fixing #426

### LINKED ISSUE #426: [Bug] - Incorrect menu entry highlighting when multiple menus use same MenuOption
### Bug
In [menu_multiple.cpp](https://github.com/ArthurSonzogni/FTXUI/blob/master/examples/component/menu_multiple.cpp) navigating between menus works as expected.

However, if you change lines 43-45 to initialize each `Menu` in the initializer list with the **same** `MenuOption`
```cpp
  // This is buggy
  MenuOption opt;
  
  auto menu_global = Container::Vertical(
      {
          Window("Menu 1", Menu(&menu_entries[0], &menu_selected[0], &opt)),
          Window("Menu 2", Menu(&menu_entries[1], &menu_selected[1], &opt)),
          Window("Menu 3", Menu(&menu_entries[2], &menu_selected[2], &opt)),
      },
      &menu_selected_global);
```
Navigating between menus will result in the **highlighting** and the `>`  to not be in sync until a subsequent keypress.
I've attached a video of this behavior [here](https://user-images.githubusercontent.com/72773002/175837975-58fb3e11-bfc9-4b29-a779-7b4b58c0e546.mp4).

### A workaround
Using different `opt`s works fine, but I would expect that I would be able to share one `MenuOption` between multiple menus:
```cpp

  // This works
  MenuOption opt;
  MenuOption opt2;
  MenuOption opt3;
  
  auto menu_global = Container::Vertical(
      {
          Window("Menu 1", Menu(&menu_entries[0], &menu_selected[0], &opt)),
          Window("Menu 2", Menu(&menu_entries[1], &menu_selected[1], &opt2)),
          Window("Menu 3", Menu(&menu_entries[2], &menu_selected[2], &opt3)),
      },
      &menu_selected_global);

```

avighnac:
# Fixes #426 

In this pull request, I've separated the focused entry option from the `MenuOptions` class (as this is buggy, one should be able to use one MenuOptions for multiple menus) and have tested my changes and this does indeed fix #426 

I have also updated the tests accordingly. Please tell if if I need to make any modifications.


ArthurSonzogni:
Thanks! PR are like this are really appreciated!

I think I would like to avoid adding new arguments. I don't think anybody is going to need this function.

Please let me some time to think about it.

One idea I have was to make this binding to work single direction. That is to say, keeping this state internal to the component, and only reflect its value to the params. This way, the component would work independently of the reflected value. You would lose the ability to control it from outside, but I believe nobody really needs it.

> as this is buggy, one should be able to use one MenuOptions for multiple menus.

That's debattable. I would rather folks to produce them from a function and have a different instance assigned to each.

However, I agree the current behavior could surprise authors.

avighnac:
> I think I would like to avoid adding new arguments. I don't think anybody is going to need this function.

I have made it so that the new argument is a default one, so the function can still be called without specifying that new argument.

> One idea I have was to make this binding to work single direction. That is to say, keeping this state internal to the component, and only reflect its value to the params. This way, the component would work independently of the reflected value. You would lose the ability to control it from outside, but I believe nobody really needs it.

If I'm interpreting this right, that's similar to what I've done! Except you can modify the value as well. If no value is specified, then a (default set to 0) internal value is maintained.

> Please let me some time to think about it.

Sure! My exams begin from tomorrow so I may not be able to check up on this very often, but you can suggest changes and I'll look at them whenever I'm free.

ArthurSonzogni:
Hi @avighnac 
I removed two way binding for Menu and Radiobox. I am not realizing we would have to do the same for Input and other component. Being able to bind them both way is something I would like to keep for Input. I am not really happy to be inconsistent.

If folks want to use the same Option for multiple element, they can just pass it by copy instead of by value.

So, I think I am not going to merge this PR. I think I prefer current behavior.

avighnac:
Hello, I just wanted to understand your perspective on the proposed changes. From what I can gather, it seems like the changes maintain compatibility. However, I'm curious about why you prefer removing flexibility by disallowing the option to be changed both ways. Additionally, I'd like to understand the reasoning behind not wanting users to be able to use the same menu option without passing by copy.

ArthurSonzogni:
Hi @avighnac,

I need to balance different problems about the API:
1. Referencing the same Option [^1] is currently not working.
2. There are many properties to expose. We may want to expose them in the function signature. This is eiter inconsistent (some properties aren't part some signature) or bloated (we maintain many variations).
3. We can decide to avoid binding both way, but we lose some potentially useful functionality.

You did (2), I tried to do (3). I believe both are worse than the status-co.


To avoid problem (1), developers just have to change:
```cpp
Menu(entries, &selected_entry, &options)
```
into
```cpp
Menu(entries, &selected_entry, options)
```
e.g. passing the option by copy instead of by reference.
That's not too hard IMO.

To avoid developers using it wrongly, maybe we can remove the Ref? That is to say replacing:
```cpp
Component Menu(ConstStringListRef entries,
               int* selected_,
               Ref<MenuOption> options = MenuOption::Vertical());
```
by:
```cpp
Component Menu(ConstStringListRef entries,
               int* selected_,
               MenuOption options = MenuOption::Vertical());
```
WDYT?

[^1]: {Menu, Radiobox, ResizeableSplit, Slider, Input}Option

avighnac:
> There are many properties to expose. We may want to expose them in the function signature. This is eiter inconsistent (some properties aren't part some signature) or bloated (we maintain many variations).

See, from my perspective, the focused entry thing is menu specific, and thus shouldn't be part of MenuOptions (if it's intended to be shared between menus). And binding it one way, as you stated, loses functionality, which I also don't like! Bloating the menu options seems like the best choice to me.

By the way, I don't see it as inconsistent either. This is part of the signature because it's menu specific, but other parts of MenuOptions aren't menu specific.

> To avoid developers using it wrongly, maybe we can remove the Ref? That is to say replacing:
Component Menu(ConstStringListRef entries,
               int* selected_,
               Ref<MenuOption> options = MenuOption::Vertical());
by:
Component Menu(ConstStringListRef entries,
               int* selected_,
               MenuOption options = MenuOption::Vertical());

Correct me if I'm wrong because it seems like I'm missing something but won't this mean that users won't be able to access the modified menu options (when FTXUI inevitably does modify them)?

PR #677, opened 2023-06-14
Problem: components can't distinguish custom events

yrashk:
They all look the same, which means they would have to check or poll all potential sources of change.

Solution: make custom events take an integer tag

I've kept the backwards compatibility by using `-1` for the default custom event (`Event::Custom`).

---

Discussion: https://github.com/ArthurSonzogni/FTXUI/discussions/676

### DIFF
--- a/include/ftxui/component/event.hpp
+++ b/include/ftxui/component/event.hpp
@@ -31,6 +31,7 @@ struct Event {
   static Event Special(std::string);
   static Event Mouse(std::string, Mouse mouse);
   static Event CursorReporting(std::string, int x, int y);
+  static Event CustomTagged(int tag);
 
   // --- Arrow ---
   static const Event ArrowLeft;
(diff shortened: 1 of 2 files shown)


ArthurSonzogni:
Hello @yrashk,
Thanks for your support!

I was thinking developers would be able to use:
```cpp
Event Event::Special(std::string input)
```
directly. For instance: `Event::Special("My event identifier")`.

Retrospectively, `Special` is not a very good name, and adding `Event::Custom` is probably a bit misleading here.

What do you think. Does this solve your need?

Note: To make your patch to work, `Event::input_` should be populate so that the `operator==` works correctly.

yrashk:
Thank you! I was thinking of `int` to clarify that the comparisons are fast and cheap. Strings may also get abused to generate runtime identifiers.

ArthurSonzogni:
Thanks @yrashk, and sorry for the very long delay. Since this PR, the recommended way to notify a specific part of the UI from another thread is `App::Post(Closure)`: the closure runs on the UI thread and can update exactly the state it's about, carrying data along, with no tag routing through every component. If you need event identity, `Event::Special("my-id")` gives distinct, comparable events. Closing for now, but feel free to reopen if a use case isn't covered.

Issue #685, opened 2023-06-22
F1-F5 creates double events in linux CTRL-ALT-Fx virtual terminals


StefanRvO:
I am unable to capture F1-F5 events if running FTXUI in the linux virtual terminals which is accessed with the CTRL-ALT-Fx modifiers. The terminals reports 'linux' as TERM.

Instead of a single event, two event are registered with the following values:
F1 -> Special({27 91 91}) + Event::Character("A")
F2 -> Special({27 91 91}) + Event::Character("B")
F3 -> Special({27 91 91}) + Event::Character("C")
F4 -> Special({27 91 91}) + Event::Character("D")
F5 -> Special({27 91 91}) + Event::Character("E")

it works correctly when running in e.g. gnome-terminal or in tmux.

Steps to reproduce:
- compile the example print_key_press
- Switch to virtual terminal with e.g. CTRL-ALT-F3
- run the print_key_press
- Press the F1-F5 buttons

Reproduced on Arch Linux, Debian 8 (32 bit) and Debian 11 (64 bit)

I'm using FTXUI v4.1.1

ArthurSonzogni:
Thanks!
I can reproduce.

So we get:
terminal | code | string
-- | -- | --
virtual terminal | 27 91 91 97 | \E[[a
gnome-terminal | 27 79 80 | \E0P

It is weird. I don't see any any standard that could apply for \E[[A in:
https://invisible-island.net/xterm/xterm-function-keys.html

I will take a look wether we can improve this.

PR #687, opened 2023-06-24
Support F1-F5 from OS terminal

ArthurSonzogni:
Bug:https://github.com/ArthurSonzogni/FTXUI/issues/685


ArthurSonzogni:
Hello @StefanRvO,
Could you please take a look? Does it looks good to you?

StefanRvO:
Yes, that seems to work. Thank you!

PR #689, opened 2023-06-25
Support F1-F4 from Putty

StefanRvO:
Putty sends different escape sequences for F1-F4 keys than
other terminals do, so translate these variations.


ArthurSonzogni:
Thanks!

I added tests and support for every possible standard. We should be good after that with regards to FN keys.

StefanRvO:
Great, crazy that there is so many different standards for the FN keys.

Issue #682, opened 2023-06-16
Draggable components.


ArthurSonzogni:
Suggested by @yrashk in discussion:
https://github.com/ArthurSonzogni/FTXUI/discussions/680

We woud like to wrap some component inside some kind of `Window`. The window
should:
- Be Draggeable using the mouse.
- (Optionally) Have a title bar
- (Optionally) Have a close button

The `WindowContainer` should allow to stack multiple `Window` on top of each
others. Clicking on a window should bring it to the top.

If possible we should find a way to make it possible to control the window
by the keyboard too.

API proposal:
```cpp
// Minimal window filled with default values:
auto win_0 = Window({
    .inner = component_1,
    .title = "Minimal example",
});

// Example with parameters with default value:
auto win_1 = Window({
    .inner = component_1,
    .title = "Example initial values",
    .left = 10,
    .right = 10,
    .width = 40,
    .height = 20,
    .closeable = false,
    .closed = false,
});

// Example with some parameters passed by reference:
int component_2_left = 10;
int component_2_right = 10;
int component_2_width = 40;
int component_2_height = 20;
auto win_2 = Window({
    .inner = component_2,
    .title = "Example with external values referenced from outside",
    .left = &component_2_left,
    .right = &component_2_right,
    .width = &component_2_width,
    .height = &component_2_height,
    .closeable = true,
    .closed = false,
};

auto win_container = WindowContainer({
  win_0,
  win_1,
  win_2,
});
```

This might be a little too "opinionated" for this to be part of FTXUI. At first,
it might be some kind of official extensions.

yrashk:
Thank you for opening an issue. Do you think we can break this down to some primitives to make any element "tell" FTXUI where it wants to be rendered (maybe a coordinates offset?)

yrashk:
As for the window component, it is definitely opinionated and should likely be external to the library. That being said, I still think that it would be essential to have primitives to build this kind of functionality. 

For example, it can be the ability for elements to tell their coordinates (or offsets), to have a `draggable` component that wraps another component and knows how to be dragged, etc.

ArthurSonzogni:
> Thank you for opening an issue. Do you think we can break this down to some primitives to make any element "tell" FTXUI where it wants to be rendered (maybe a coordinates offset?)

It is already available via the `reflect` element. See this reply: https://github.com/ArthurSonzogni/FTXUI/issues/521#issuecomment-1336250274

yrashk:
[...]

PR #690, opened 2023-06-25
Feature: Windows.

ArthurSonzogni:
Into ftxui/component/, add:
```cpp
Container::Stacked(...)
Window(...);
```

Together, they can be used to display draggable/resizable windows.

Bug:https://github.com/ArthurSonzogni/FTXUI/issues/682


ArthurSonzogni:
@yrashk, I believe I am now happy with the API and the implementation.

Could you please take a look? Is there any names / API to be improved here before merging this patch?

See current state:
https://www.youtube.com/watch?v=4rR7He6_3Uc

Current API:
```cpp
Component Stacked(Components children);
```
```cpp
Component Window(WindowOptions option);
```
For instance:
```cpp
auto window_1 = Window({
  .inner = wrapped_component,
  .title = "title"
})

auto windows = Container::Stacked({
  window_1,
  [...]
})

```


This can be customized with:
```cpp
// @brief Option for the `Window` component.
// @ingroup component
struct WindowOptions {
  Component inner;            /// < The component wrapped by this window.
  ConstStringRef title = "";  /// < The title displayed by this window.

  Ref<int> left = 0;     /// < The left side position of the window.
  Ref<int> top = 0;      /// < The top side position of the window.
  Ref<int> width = 20;   /// < The width of the window.
  Ref<int> height = 10;  /// < The height of the window.

  Ref<bool> resize_left = true;   /// < Can the left side be resized?
  Ref<bool> resize_right = true;  /// < Can the right side be resized?
  Ref<bool> resize_top = true;    /// < Can the top side be resized?
  Ref<bool> resize_down = true;   /// < Can the down side be resized?

  /// An optional function to customize how the window looks like:
  std::function<Element(const WindowRenderState&)> render;
};
```

and you can even customized how to render the window:
```cpp
// Parameter pack used by `WindowOptions::render`.
struct WindowRenderState {
  Element inner;             /// < The element wrapped inside this window.
  const std::string& title;  /// < The title of the window.
  bool active;               /// < Whether the entry is the active one.
  bool drag = false;         /// < Whether the window is being dragged.
  bool resize = false;       /// < Whether the window is being resized.
  bool hover_left = false;   /// < Whether the resizeable left side is hovered.
  bool hover_right = false;  /// < Whether the resizeable right side is hovered.
  bool hover_top = false;    /// < Whether the resizeable top side is hovered.
  bool hover_down = false;   /// < Whether the resizeable down side is hovered.
};
```

yrashk:
I think it's a great start, indeed! Thank you so much for putting this work.

The first thing that caught my eye (though I am not sure I got this right): moving `Components` into `Stacked` means that we can't add/remove children from this component dynamically?

ArthurSonzogni:
Thanks!

> The first thing that caught my eye (though I am not sure I got this right): moving Components into Stacked means that we can't add/remove children from this component dynamically?

The container stays a ComponentBase, so you can continue to add/remove children dynamically like in any other components.

PR #691, opened 2023-06-27
Improve unicode codepoint Bisearch performance

StefanRvO:
Improve the performance of the functions for searching for codepoints in a table by passing the table array in as a reference instead of copying it.

This doesn't really improve the results of the benchmarks for some reason, but in my application with GCC8 i686, Intel Atom E3950 and a 180x60 terminal, this reduces the time for rendering a frame from ~31 ms to ~5.5 ms.


ArthurSonzogni:
Wow! I can't believe this got unnoticed for so long. Thank you!

I added a benchmark. I measured this improves debug performance by 3x and Release performance by 0x. I guess the compiler was able to optimize this function.

ArthurSonzogni (approved):
LGTM!

StefanRvO:
Yeah, I'm guessing my older GCC8 for some reason isn't able to optimize it, but never compilers probably are. I would guess that the compiler would see these as great in-lining candidates as it is only used inside the module, but at least GCC 8 with -O3 doesn't do that.

Issue #692, opened 2023-06-30
Unused variable warning in 'screen_interactive.cpp' for _WIN32 builds


cblack-dev:
Building ftxui-component project in VS2022 (Community)
Windows SDK Version: 10.0.22000.0
Platform Toolset: LLVM (clang-cl)
C++ Language Standard: ISO C++20 Standard (/std:c++20)
Warning Level: Level 3 (/W3) /WX

"..src\ftxui\component\screen_interactive.cpp(73,15): error : unused variable 'timeout_microseconds' [-Werror,-Wunused-const-variable]"
Since the variable 'timeout_microseconds' is not referenced in _WIN32 builds [it is referenced for the else cases of __EMSCRIPTEN__ or the fall-through to POSIX (Linux & Mac)] it triggers the above warning.

Possible solutions:
1. If C++17 support is expected, timeout_microseconds can be marked [[maybe_unused]].
2. timeout_microseconds can be moved out of global build scope and into the two branches it is used.

ArthurSonzogni:
Thanks! I guess this is now fixed.

PR #693, opened 2023-06-30
Resolve unused varaible warning in _WIN32 builds of screen_interactiv…

cblack-dev:
Bug:#692

Marks constexpr int timeout_microseconds on line 73 as [[maybe_unused]] to resolve unused variable warning on _WIN32 builds.


ArthurSonzogni (approved):
Thanks! This looks good to me!

PR #701, opened 2023-07-25
Improve const correctness in the Screen class

### LINKED ISSUE #700: Should Screen class have const accessors?
the `Screen` class has accessors for the pixel color and methods like `ToString()`, which are not marked as `const`. Is there a reason why they are not? I'd say that there should be these methods (among other ones like them):
```cpp
const Pixel& PixelAt(int x, int y) const;

std::string ToString() const;
```

Am I missing something?

niosus:
This PR addresses what we discussed in #700 by making the following changes:
- Add a const variant to the accessor functions
- Make Print and ToString functions const
- Move Doxygen comments to the header (I am happy to remove this change but I decided to kill two birds with one stone here)

This, when merged, should close #700

### DIFF
--- a/include/ftxui/screen/screen.hpp
+++ b/include/ftxui/screen/screen.hpp
@@ -66,13 +66,18 @@ class Screen {
   static Screen Create(Dimensions dimension);
   static Screen Create(Dimensions width, Dimensions height);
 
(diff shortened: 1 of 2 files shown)


ArthurSonzogni:
Thanks!

PR #704, opened 2023-07-27
Performance improvement by refactoring pixel styles

clement-roblot:
Hi,

I described the idea as part of #703, but turns out I was looking at version 4.0.0 and not main so a few things changed in what I described. As for the performance gains, I am using a lightly modified Downloader example and using perf to measure. I went from using 12.03% of the processing just for `UpdatePixelStyle` to 6.97% (measured over 3 runs each time).

A few points of interest:
- I kept `automerge` outside of the style bitfield because it is used independently from the other style properties.
- I didn´t implement any tests, would like to though. To do that I would want to move the function UpdatePixelStyle into a method of the Screen class. If that's ok, then I do that right away.

Closes #703

PR #708, opened 2023-07-29
Added the print stage to the benchmarks

clement-roblot:
The benchmarks were only doing a render but not calling the ToString() method (as done in the Screen:Print method). I added this.

This allows us to see the performance improvement in https://github.com/ArthurSonzogni/FTXUI/pull/704.


ArthurSonzogni:
It makes sense! Thank you!

PR #709, opened 2023-07-29
Upgrade the version of google benchmark

clement-roblot:
This upgrades the version of google benchmark used. The main reason for this upgrade is that the old version doesn't display the `OVERALL_GEOMEAN` when comparing two benchmarks.


clement-roblot:
Question @ArthurSonzogni : why specify the commit hash? Is it by "fear" to tag could be moved by someone in the future?

ArthurSonzogni:
Thanks you!

> Question @ArthurSonzogni : why specify the commit hash? Is it by "fear" to tag could be moved by someone in the future?

Yes, that's the reason. I fully trust folks from google-benchmark repository, but I prefer to be consistent whatever repository I fetch from.

PR #712, opened 2023-07-30
Added rw-tui to the list of projects that are using FTXUI

LeeKyuHyuk:
Since I developed `rw-tui` using FTXUI, I want to add it to the `README.md`.

This utility was created as a TUI application for Linux with reference to RWEverything.

rw-tui

Example of using rw-tui

**Features supported:**
- Output the memory area as a table and change the value
- Output configuration space MMIO area of PCI device as table
- PCI configuration space output


ArthurSonzogni:
Beautiful! Thanks for using FTXUI and adding your project!

PR #719, opened 2023-08-02
Possible clarification and reimplementation of Ref<T> classes

mr-mocap:
I was a bit confused by the current implementation of the ConstRef<T> and Ref<T> classes, so I decided to see if I could clarify your original intent.

I hated to "mess up" the simplicity of the original classes, but I think I've codified what you were trying to do.

I'm hoping you will find this useful enough to merge into your repo.

### DIFF
--- a/include/ftxui/util/ref.hpp
+++ b/include/ftxui/util/ref.hpp
@@ -3,6 +3,7 @@
 
 #include <ftxui/screen/string.hpp>
 #include <string>
+#include <variant>
 
 namespace ftxui {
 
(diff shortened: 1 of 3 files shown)


mr-mocap:
Excellent!

I was going to make some more changes to the StringRef and ConstStringRef classes, mainly seeing if I could make it a variant of a std::string and a std::string_view, as this also seems to be in the spirit of either wanting to initialize with a std::string or a hard-coded const char * string ("like this").  In this way, only an immutable string_view would mostly be passed around, which is really lightweight.

Don't worry about immediate responses, as I prefer quality over most other things.  This may take some back and forth to really get it polished.

ArthurSonzogni:
- I removed the c++20 check in example, because we are already using it.
- I added operator= for Ref.
- I removed some std::move when clang-tidy suggested it was not useful.
- I simplified some things.

Does it looks good to you?

mr-mocap:
It looks good so far.
I have changed the implementation of ConstStringRef on my side, but I cannot figure out how to get that change into this pull request.
Perhaps I should just make another pull request after this closes.

mr-mocap:
Found out how!
Try this out...

ArthurSonzogni:
Thanks!
I think you overrode my last change. I will push again what I did, and if you think there are something to tweak, I will let you publish another PR.

PR #726, opened 2023-08-16
Add a DisableMouseEvents() on ScreenInteractive to disable mouse events

StefanRvO:
When mouse events are enabled, it is not possible to select text in the terminal and copy it somewhere else. This could be usefull for some applications if they don't need to handle mouse events.

Add a function on the ScreenInteractive class to disable grabbing of mouse events so that it is e.g. possible to select text in the user interface. The function needs to be called on the screen object before starting the application loop if such a behaviour is desired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ current (development)
   Component Slider(SliderOption<T> options);
   Component ResizableSplit(ResizableSplitOption options);
   ```
+- Feature: Add `ScreenInteractive::TrackMouse(false)` disable mouse support.
 
 ### Dom
 - Feature: Add `hyperlink` decorator. For instance:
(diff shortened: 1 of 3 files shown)


ArthurSonzogni:
Thanks I like it!

I will take a look tomorrow, tweak a few things and merge this patch.

eliasdorneles:
> When mouse events are enabled, it is not possible to select text in the terminal and copy it somewhere else. 

Isn't it just that to use the graphical mouse events you need to press a modifier key?

In my case, I'm still able to select text if I press <kbd>Shift</kbd> at the same time.

Check this item from the FAQ of the Textual Python library, I believe it also applies here: https://github.com/Textualize/textual/blob/main/FAQ.md#how-can-i-select-and-copy-text-in-a-textual-app

StefanRvO:
Hmm, you are right, i didn't know that.
Still think it is maybe worth it to at this so stuff can be marked without the modifier key? It is very little code and makes things a bit easier.

ArthurSonzogni:
Thanks! I merged the PR with some tweaks to it.

PR #727, opened 2023-08-16
Fix Wshadow warning in Color::Interpolate

StefanRvO:
-Wshadow would warn on shadowed local variables
in the Color::Interpolate function


ArthurSonzogni:
Thanks!

PR #728, opened 2023-08-16
Add -Wuseless-cast to FTXUI_DEV_WARNINGS

StefanRvO:
Add the -Wuseless-cast to the FTXUI_DEV_WARNINGS and fix the compiler complaints about useless casts


StefanRvO:
Hmm, seems like clang doesn't have this warning :/

StefanRvO:
Changed it so the flag is only set if using GCC

ArthurSonzogni:
Thanks!

PR #729, opened 2023-08-16
Add -Wold-style-cast to FTXUI_DEV_WARNINGS

StefanRvO:
Add the -Wold-style-cast to the FTXUI_DEV_WARNINGS and fix the compiler complaints about C casts


ArthurSonzogni:
Thank you!

I like old style cast. So sorry, I would like not to merge this PR. Is there any reason embedders of FTXUI might need this?

StefanRvO:
Ok, that is fair. Not really, other than if a codebase FTXUI is compiled into sets the -Wold-style-cast warning as default, as mine do, so i have to except that warning for FTXUI when i add it. But it is fine, not much of a hassle :)

PR #730, opened 2023-08-16
Tidy up some ConstRef<T> and Ref<T> constructor cases

mr-mocap:
Here is a slight improvement in the handling of the construction of those two classes.

It uses a pass-by-value-when-moving type of approach.  This will handle both L-values and R-values cleanly.  Of course, this assumes that moves are free (which they typically are) and are inlined.

What do you think?

P.S. Yes, I meant "reseatable" and not "resetable".  :)


ArthurSonzogni:
Thanks!
This looks good to me!
Let's see what the CQ says, I will merge the patch this week-end.

ArthurSonzogni:
Thanks! PR merged!

PR #731, opened 2023-08-17
Dropdown now closes when using return key to select

mingsheng13:
Hi, this is my first time doing pull request! I think it would be nice if the dropdown menu closes after selecting an entry. Please tell me if I'm doing anything wrong. Thank you!

### DIFF
--- a/src/ftxui/component/dropdown.cpp
+++ b/src/ftxui/component/dropdown.cpp
@@ -58,16 +58,22 @@ Component Dropdown(ConstStringListRef entries, int* selected) {
       });
     }
 
-    bool OnEvent(ftxui::Event event) override
-    {
-      if (event == Event::Return && show_)
-      {
-        children_[0]->OnEvent(event);
+    // Switch focus in between the checkbox and the radiobox when selecting it.
+    bool OnEvent(ftxui::Event event) override {
+      const bool show_old = show_;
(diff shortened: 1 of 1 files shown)


ArthurSonzogni:
Thanks!

I tweaked it a bit to support the mouse. I merged the PR. Thanks!

src/ftxui/component/dropdown.cpp
```diff
@@ -58,16 +58,22 @@ Component Dropdown(ConstStringListRef entries, int* selected) {
...
-      if (event == Event::Return && show_)
-      {
-        children_[0]->OnEvent(event);
+    // Switch focus in between the checkbox and the radiobox when selecting it.
+    bool OnEvent(ftxui::Event event) override {
+      const bool show_old = show_;
+      const int selected_old = *selected_;
+      const bool handled = ComponentBase::OnEvent(event);
+
+      if (!show_old && show_) {
```
mingsheng13, line 67: I don't understand this line... why is this required?

ArthurSonzogni: That's when the dropdown is activated, we put the focuse to the list. It avoid the user to press a key

mingsheng13: and what's the const bool show_old for?

PR #735, opened 2023-08-27
input can now use overwrite mode when toggled with insert key

mingsheng13:


mingsheng13:
overwrite

ArthurSonzogni:
Thank you very much! This looks very nice!

include/ftxui/component/component_options.hpp
```diff
@@ -177,6 +177,7 @@ struct InputOption {
   std::function<Element(InputState)> transform;
   Ref<bool> password = false;  /// < Obscure the input content using '*'.
   Ref<bool> multiline = true;  /// < Whether the input can be multiline.
+  Ref<bool> overwrite = false; /// < Overwrite text at cursor.
```
ArthurSonzogni, line 180: I think developers are more aware of the "insert" mode than the "overtype" one.

What about using:
```cpp
  Ref<bool> insert = true;     /// < Insert or overtype character mode.
```

src/ftxui/component/input.cpp
```diff
@@ -101,7 +101,8 @@ class InputBase : public ComponentBase, public InputOption {
   Element Render() override {
     const bool is_focused = Focused();
     const auto focused =
```
ArthurSonzogni, line 103: I think it would be easier to read this way:
```cpp
    const auto focused = (!is_focused && !hovered_) ? select
                         : overwrite()              ? focusCursorBlockBlinking()
                                                    : focusCursorBarBlinking();

```

ArthurSonzogni (changes requested):
Thanks! Here are some comments:

After they are addressed, I will add a patch to add some test and update the CHANGELOG.

mingsheng13:
@ArthurSonzogni I've changed the codes. Could you please have a look :)

mingsheng13 (commented):
done

ArthurSonzogni (approved):

Issue #672, opened 2023-06-09
[feature request] IP Input component


mohsenomidi:
@ArthurSonzogni 
Thanks for your amazing library 

I am wondering to know, is there any plan to add ip address input component separated by .

Thanks 🙏

ArthurSonzogni:
Hi @mohsenomidi 

This is quite specific, and there are many exotic ways to write an IP address. Not sure if this is something easy to do perfectly.

Ideally, I would like to reference some "component" libraries for FTXUI. An IP input one would be great:
https://github.com/ArthurSonzogni/FTXUI/issues/672

If you have a good idea how it should behave maybe you could share an implementation?

mohsenomidi:
Thanks for your reply, I will thinking about it, and will share if I find the way

One more thing, it seems, the link you provided is not referenced to the mentioned component

ArthurSonzogni:
The correct link is:
https://github.com/ArthurSonzogni/FTXUI#libraries-for-ftxui

mingsheng13:
Hi, if I were to implement one, what's the way that you prefer? Should I add an InputOption such as InputOption::IP? Or a new cpp file: IpInput.cpp? Thanks.

mohsenomidi:
Completed 👍

PR #737, opened 2023-08-28
Ip input component

mingsheng13:
I wrote a component for inputting IPV4 address. One bug that I wasn't able to fix is when using cursor to remove the dots it breaks the logic. I think the only way to fix this is to be able to get the cursor_position() outside the Input component.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,6 +119,7 @@ add_library(component
   src/ftxui/component/event.cpp
   src/ftxui/component/hoverable.cpp
   src/ftxui/component/input.cpp
+  src/ftxui/component/ip_input.cpp
   src/ftxui/component/loop.cpp
   src/ftxui/component/maybe.cpp
   src/ftxui/component/menu.cpp
--- a/examples/component/input.cpp
+++ b/examples/component/input.cpp
@@ -20,7 +20,7 @@ int main() {
   std::string password;
 
   Component input_first_name = Input(&first_name, "first name");
-  Component input_last_name = Input(&last_name, "last name");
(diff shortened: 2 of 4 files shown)


ArthurSonzogni:
Thanks @mingsheng13 !

As as said in: https://github.com/ArthurSonzogni/FTXUI/issues/672
I believe an IpInput component to be too specific for FTXUI. If you made a repository, I could contribute to make the CMake config usable for others and reference it from:
https://github.com/ArthurSonzogni/FTXUI#libraries-for-ftxui
?

I really would like folks to add their own set of components, maybe this could be the first one?

mingsheng13:
> As as said in: #672 I believe an IpInput component to be too specific for FTXUI. If you made a repository, I could contribute to make the CMake config usable for others and reference it from: https://github.com/ArthurSonzogni/FTXUI#libraries-for-ftxui ?


Thank you for the reply. That's actually what I wanted to do! I'll make it in my repo.

mingsheng13:
I've moved it to here. https://github.com/mingsheng13/ip-input-ftxui . Please tell me if I'm doing anything wrong. Thanks :)

ArthurSonzogni:
Awesome!

I [forked your repository](https://github.com/ArthurSonzogni/ftxui-ip-input) and improved it. I will work on it for a few days and submit PR against your repository when I am done.

mingsheng13:
Thanks! I had a look at the fork and was impressed by the code quality. Wish I could be as good as you 😭

ArthurSonzogni:
Thanks!

I added `ftxui-ip-input` in the FTXUI README.

PR #738, opened 2023-08-28
Fix CMAKE_UNITY_BUILD usage

ClausKlein:
Add missing include guard.

This fix https://github.com/cpp-best-practices/cmake_template/issues/64


ArthurSonzogni:
Thanks! I remember Chromium has some [options](https://chromium.googlesource.com/chromium/src.git/+/65.0.3283.0/docs/jumbo.md) for this, I  couldn't imagine cmake has options like this.

I updated the patch to avoid pragma, added comments, and updated the CHANGELOG.

PR #742, opened 2023-08-30
a grid container

mingsheng13:
Hi, I've written a simple grid container using Horizontal and Vertical Containers. But I think this should be put into the container.cpp instead of a different library. Which one would you prefer? Thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,7 +294,7 @@ Prebuilt components are declared in [<ftxui/component/component.hpp>](https://ar
 
 ## Libraries for FTXUI
 - *Want to share a useful component using FTXUI? Feel free adding yours here*
-
+- [GridContainer](https://github.com/mingsheng13/grid-container-ftxui)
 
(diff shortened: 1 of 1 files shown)


ArthurSonzogni:
Thanks! I like you implementation! Sharing the same variable for the column/row was smart.

I always wanted to add it to FTXUI. I might want to do something a little bit more customized in order to have the column/row number be set by a variable, and support user to push new children.

mingsheng13:
Glad you liked it! I was wondering how can I use external libraries like this one in FTXUI? How are you going to link it?

ArthurSonzogni:
> Glad you liked it! I was wondering how can I use external libraries like this one in FTXUI? How are you going to link it?

I puished some CMakeLists configuration to `ip-input-ftxui` and `grid-container-ftxui` and a README to document how to do it.

What do you think about them?

PR #751, opened 2023-09-20
Example: fix small flaw in example/button.cpp

jiansyuan:
I noticed that there are 2 buttons labeled '-10', but one of them increases by 10.


ArthurSonzogni (approved):
Thanks!

PR #755, opened 2023-10-01
Scrollbar coloring

### LINKED ISSUE #754: scroll indicator not colorable?
Hi there,

I'm using a vscroll_indicator in an app where I change the color of every element using the decorators color and bgcolor. Those work well on everything that I throw at them except the vscroll_indicator.

I can modify the vscroll_indicator implementation to change the colors in the render but that's definitely a dirty solution. I'm a bit confused as to why/how the color and bgcolor decorators can catch all the pixels in the screen except the ones from the vscroll_indicator :thinking: .

Here is the example button_in_frame.cpp with a custom color scheme that applies to every element except the scroll bar:

```
// Copyright 2022 Arthur Sonzogni. All rights reserved.
// Use of this source code is governed by the MIT license that can be found in
// the LICENSE file.
#include <memory>  // for allocator, __shared_ptr_access, shared_ptr
#include <string>  // for to_string, operator+

#include "ftxui/component/captured_mouse.hpp"  // for ftxui
#include "ftxui/component/component.hpp"       // for Button, Renderer, Vertical
#include "ftxui/component/component_base.hpp"  // for ComponentBase
#include "ftxui/component/component_options.hpp"   // for ButtonOption
#include "ftxui/component/screen_interactive.hpp"  // for ScreenInteractive
#include "ftxui/dom/elements.hpp"  // for operator|, text, Element, hbox, separator, size, vbox, border, frame, vscroll_indicator, HEIGHT, LESS_THAN
#include "ftxui/screen/color.hpp"  // for Color, Color::Default, Color::GrayDark, Color::White

using namespace ftxui;

int main() {
  int counter = 0;
  auto on_click = [&] { counter++; };

  Color foregroundColor = Color::RGB(253, 202, 85);
  Color backgroundColor = Color::RGB(63, 41, 30);

  auto style = ButtonOption::Animated(backgroundColor, foregroundColor,
                                      backgroundColor, foregroundColor);

  auto container = Container::Vertical({});
  for (int i = 0; i < 30; ++i) {
    auto button = Button("Button " + std::to_string(i), on_click, style);
    container->Add(button);
  }

  auto renderer = Renderer(container, [&] {
    return vbox({
               hbox({
                   text("Counter:"),
                   text(std::to_string(counter)),
               }),
               separator(),
               container->Render()| vscroll_indicator | frame |
                   size(HEIGHT, LESS_THAN, 20),
           }) |
           border | color(Color::RGB(253, 202, 85)) | bgcolor(Color::RGB(63, 41, 30));
  });

  auto screen = ScreenInteractive::FitComponent();
  screen.Loop(renderer);

  return 0;
}
```

clement-roblot:
Hi there,

This a proposed MR to fix #754. While building the scroll bar the pixels were completely reseted thus canceling any style previously applied to said pixels. This MR removes this resetting of the pixels and leaves only the drawing of the shape of the scroll bar.

### DIFF
--- a/src/ftxui/dom/scroll_indicator.cpp
+++ b/src/ftxui/dom/scroll_indicator.cpp
@@ -64,7 +64,6 @@ Element vscroll_indicator(Element child) {
         const bool down = (start_y <= y_down) && (y_down <= start_y + size);
 
         const char* c = up ? (down ? "┃" : "╹") : (down ? "╻" : " ");  // NOLINT
-        screen.PixelAt(x, y) = Pixel();
         screen.PixelAt(x, y).character = c;
       }
     }
@@ -121,7 +120,6 @@ Element hscroll_indicator(Element child) {
 
         const char* c =
             left ? (right ? "─" : "╴") : (right ? "╶" : " ");  // NOLINT
-        screen.PixelAt(x, y) = Pixel();
         screen.PixelAt(x, y).character = c;
       }
     }
--- a/src/ftxui/dom/scroll_indicator_test.cpp
+++ b/src/ftxui/dom/scroll_indicator_test.cpp
@@ -136,7 +136,7 @@ TEST(ScrollIndicator, VerticalColorable) {
   //           "│1  ┃│\r\n"
   //           "│2   │\r\n"
   //           "│3   │\r\n"
-  //           "╰────╯");
+  //           "╰────╯"
 
   auto element = MakeVerticalList(0, 10) | color(Color::Red);
   Screen screen(6, 6);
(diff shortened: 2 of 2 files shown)


clement-roblot:
Had a doubt and looked a bit more in the details of a pixel. Although resetting the colors doesn´t make sense here, but resetting other fields in the pixels might. For instance, do we want to keep a pixel blinking when it is converted into a scroll bar?

I feel like I should leave the resetting of the pixels as they were before this MR but implement a system to backup and re-apply some of the pixel's characteristics. From the list I feel like I want to re-apply:
- bold
- dim
- inverted
- background_color
- foreground_color

I would reset:
- blink
- underlined
- underlined_double
- strikethrough

And I don´t really know if I should reset or not :
- automerge
- hyperlink

ArthurSonzogni:
Thanks you!

I believe the current patch to be good. We will see later if someone really need to reset something.

PR #760, opened 2023-10-04
Create fireworks class

CVelez21:
Hello! This class provides a template to use the static animations and easily adjust the speed in a concise manner

### DIFF
--- a/src/ftxui/component/fireworks
+++ b/src/ftxui/component/fireworks
@@ -1,0 +1,266 @@
+#include <chrono>
+#include <thread>
+#include <iostream>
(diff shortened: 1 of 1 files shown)


CVelez21:
Approved

mavam:
Mind posting a quick GIF/video on how this looks like? 🎆

ArthurSonzogni:
Thanks!
This is a funny usage of FTXUI.

I don't believe someone would need it as a builtin component, so I am not going to accept this PR. This could still be an external component and referenced from the FTXUI README if someone believe this is useful.

PR #763, opened 2023-10-11
Added the ability to filter characters inputted in an input field

clement-roblot:
Hi there,

I wanted an input field that allows only for numbers to be inputted so I added the ability to filter what type of characters are accepted in an input field.

I made the OnEvent return false when a character is rejected. Is this how it should behave in your opinion?


clement-roblot:
Haaa, I see. Indeed your way is a lot more generic and everyone does whatever they want like this. I modified the input example to reflect this, is this what you had in mind regarding advertising in the documentation?

ArthurSonzogni:
Thank you very much!
I added a section into the documentation.

PR #768, opened 2023-10-14
fix the doxygen documentation by moving comments at the right place

benoitdudu:
By reading the documentation I found that some descriptions were not matching the class member name.
The fix is quite trivial (I moved the comment used by doxygen above the class member).

### DIFF
--- a/include/ftxui/component/component_options.hpp
+++ b/include/ftxui/component/component_options.hpp
@@ -175,9 +175,12 @@ struct InputOption {
 
   // Style:
   std::function<Element(InputState)> transform;
-  Ref<bool> password = false;  /// < Obscure the input content using '*'.
-  Ref<bool> multiline = true;  /// < Whether the input can be multiline.
-  Ref<bool> insert = true;     /// < Insert or overtype character mode.
+  /// < Obscure the input content using '*'.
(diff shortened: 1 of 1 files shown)


include/ftxui/component/component_options.hpp
```diff
@@ -175,9 +175,12 @@ struct InputOption {
 
   // Style:
   std::function<Element(InputState)> transform;
-  Ref<bool> password = false;  /// < Obscure the input content using '*'.
```
ArthurSonzogni, line 178: I believe the correct fix would be to remove the space in between the "///" and the "<".

See the `Putting documentation after members` section of:
https://www.doxygen.nl/manual/docblocks.html

Could you do this instead?

ArthurSonzogni (commented):
Thanks!

include/ftxui/component/component_options.hpp
benoitdudu: You are right. I made the change and fix the others similar problems.

ArthurSonzogni: Excellent! Thanks!

ArthurSonzogni (approved):

PR #769, opened 2023-10-17
update nix and add dev shell

Alizter:
We update the lock on the nix flake and also add a dev shell. This means you can do `nix build` to build the project and `nix develop` to drop into a development environment with cmake and clang.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,71 +8,69 @@
 
   outputs = {self, nixpkgs, flake-utils}:
     flake-utils.lib.eachDefaultSystem (system:
-    let
-      pkgs = import nixpkgs { inherit system; };
-      llvm = pkgs.llvmPackages_latest;
-    in
-      {
-        packages = rec { 
-          default = pkgs.stdenv.mkDerivation rec {
-            pname = "ftxui";
-            version = "v4.0.0";
-            src = pkgs.fetchFromGitHub {
-              owner = "ArthurSonzogni";
-              repo = "FTXUI";
-              rev = version;
-              sha256 = "sha256-3kAhHDUwzwdvHc8JZAcA14tGqa6w69qrN1JXhSxNBQY=";
-            };
-
-            nativeBuildInputs = [
-              pkgs.cmake
-            ];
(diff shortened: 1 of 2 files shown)


ArthurSonzogni:
Thank you!

I need to spend some time to understand nix again. Thanks for your patience.

Alizter:
No worries! Let me know if you have any questions.

flake.nix
```diff
@@ -8,9 +8,11 @@
 
   outputs = {self, nixpkgs, flake-utils}:
     flake-utils.lib.eachDefaultSystem (system:
-      let pkgs = import nixpkgs { inherit system; }; in
-      {
-        packages.ftxui = pkgs.stdenv.mkDerivation rec {
+    let pkgs = import nixpkgs { inherit system; }; in
+    let llvm = pkgs.llvmPackages_latest; in
+    {
+      packages = rec {
```
ArthurSonzogni, line 14: Question: I observed the construct with:
```nix
packages = rec {
  default = ...
  ftxui = default
}
```
worked better than:
```nix
packages.ftxui = ...
```

I don't understand why. Would you have an explanation?

ArthurSonzogni (approved):

ArthurSonzogni:
Thanks!

flake.nix
Alizter: Hi, good question! When you go to a project and do a literal `nix build` then nix tries to build the "default" package in the flake. The previous setup called this `ftxui` which is not really useful since there is basically only one package to build here. I therefore renamed the existing `packages.ftxui` to `packages.default`, allowing `nix build` to work without any extra arguments.

In order to not break anybody else's workflow, I added `packages.ftxui` back in as an alias of `packages.default`. In Nix, if you are defining a set where one element refers to another at the time of definition, you will need it to be a recursive definition, hence the `rec` flag.

Notice that the default devShell is called `devShells.default` which allows you to do a simple `nix develop` in the repo in order to setup a dev environment.

Hopefully that clears some things up!

Issue #773, opened 2023-10-23
Click and drag tiggers buttons and checkboxes repeatedly


clement-roblot:
Hi there,

I made a video but for some reason it doesn´t look like the real deal. But here it is anyway:

![2023-10-23 21-19-00](https://github.com/ArthurSonzogni/FTXUI/assets/2040179/d87b3392-480f-4922-ae79-4f16497a3544)

If you click on a checkbox or a button and keep your left mouse button pressed and start drawing the mouse sideways, it will toggle on and off the checkbox (or repeatedly click the button). This behavior leads to sometimes having a button triggering the onClick event twice even though I though I clicked it only once.

I would like to add an option to the button object (and also to the checkbox object) that would enable a mechanism looking for a released to press motion to trigger an OnClick event, rather than just looking for press events as it is now. Here is a quick example that works for me for a button:

![image](https://github.com/ArthurSonzogni/FTXUI/assets/2040179/6ed284a4-91bf-4d5e-b393-76cf282bdacb)

Does such an implementation look valid, should I go forth with a PR?

ArthurSonzogni:
It might. Would you like to propose a pull request?

PR #774, opened 2023-10-24
Checkbox button debounce

clement-roblot:
Hello there,

This is a suggestion to solve #773. I personally have this issue with buttons and discovered it existed also with checkboxes, I think those are the only 2 components being afflicted by this issue, would you agree?


ArthurSonzogni:
Thanks!

I believe this fix partially the problem. See this recording:

https://github.com/ArthurSonzogni/FTXUI/assets/4759106/203dd5b9-a654-4d55-ad17-01a9b9b02647



I am thinking about two alternative.

## CaptureMouse

1. Make every component catching mouse pressing to `CaptureMouse` and keep the handle until getting an event indicating the mouse moved without pressing the button.
2. Implement a "top" level handler that would `CaptureMouse` in a similar way.

(1,2) would prevent mouse event started from one component to affect a different one before the end of the movement.

## Decompose "Press/Move/Release" via FTXUI.

FTXUI would compute the state of the event. Instead of exposing "pressed/not-pressed", we would expose whether we transitionned from pressed -> not-pressed, pressed->pressed and pressed->not-pressed.

This way, checkbox and button would check the "pressed->not-pressed" state and igfnore the "pressed->pressed" one.

---


I think I would try (2) based on your CL.L

ArthurSonzogni:
Hi @clement-roblot,
I implemented the second solution I was talking above.
Could you please take a look?

clement-roblot:
OMG, this is glorious. Thanks a lot @ArthurSonzogni  :heart:

PR #776, opened 2023-11-11
Input onchange not called

chrysante:
I noticed that the `on_change` callback for the `Input()` component is not called when backspace or delete keys are pressed. This PR fixes that.


ArthurSonzogni (approved):
Thanks!

Issue #791, opened 2023-12-15
Mouse input does not work correctly on urxvt: Moving the cursor always claims the mouse button is clicked.


psychon:
Hi,

running the `print_key_press` example in `urxvt` and moving the mouse around produces e.g.
```
( 27 91 60 51 50 59 56 53 59 52 50 77 ) -> mouse_left_pressed_control(84,9)
```
Notice that this incorrectly claims that the left mouse button is pressed.

The same thing in `xterm`:
```
( 27 91 60 51 53 59 55 57 59 49 53 77 ) -> mouse_none_pressed_control(78,13)
```
This issue makes basically anything interactive unusable. E.g. button will trigger every time the cursor moves slightly across it.

ArthurSonzogni:
Hello @psychon

Thanks for reporting this 

I think this was improved ~30 days ago with:
https://github.com/ArthurSonzogni/FTXUI/pull/774

Every builtin components have been updated.
A new API for custom on is available
```cpp
  bool Mouse::IsPressed(Button button) const;
  bool Mouse::IsHeld(Button button) const;
  bool Mouse::IsReleased(Button button) const;
```

Does the new version satisfies you, or do we need something different?

psychon:
Ah, you are right. I was initially experimenting with the latest release and then switched to latest `main` before reporting this issue.

However, "things" still don't seem to work correctly. I switched to the `component/dropdown` example. My instructions:
1. click an entry (to open the dropdown)
2. move the mouse to an item
3. click
4. move the mouse

With `xterm`, the dropdown closes at step 3 and the mouse move has no effect. With urxvt, the mouse click has no effect and only at step 4 does "something" happen.

Put differently: Yes, I can now interact with the dropdown via the mouse, but things still behave weird / unexpected.

I checked out git at `c31aecf2edde81d20f4f4bc8c97af053026483a6^` (= before #774) and there things indeed behave as I described out. That's a lot worse than the current state of things.

psychon:
I found http://pod.tst.eu/http://cvs.schmorp.de/rxvt-unicode/doc/rxvt.7.pod#Mouse_Reporting and the following sub-sections. ftxui seems to enable 1000, 1003, 1015, and 1016.

I did some more tests showing the raw input that was received in a human readable form:
```diff
diff --git a/examples/component/print_key_press.cpp b/examples/component/print_key_press.cpp
index c481271..221f93a 100644
--- a/examples/component/print_key_press.cpp
[...]

PR #794, opened 2023-12-16
Support move movement.

ArthurSonzogni:
- Record whether the mouse moved. Suppress the previous alternative implementation.
- Record mouse.control modifier.

Fixed:https://github.com/ArthurSonzogni/FTXUI/issues/791


ArthurSonzogni:
Hello @psychon,
Could you please take a look? Does this improve the situation?

psychon:
> Does this improve the situation?

TL;DR: Yup, it does. I can't spot anything really wrong that can be fixed in ftxui.

Let's look at `ftxui_example_print_key_press` again:

- Open a new urxvt instance, run this, move the mouse around reports things like `mouse_none_pressed_control_shift_meta(95,24)`; this is a bug in urxvt that I found yesterday and I am still trying to report upstream (urxvt's code reports "button -1 was last pressed" and that messes up everything)
- After any mouse press, things become `mouse_left_moved(61, 6)` or `mouse_right_moved(63, 5)` or something like that (same for middle button and wheels)
- Click-and-drag produces stuff like
  -  `mouse_left_pressed(...)`
  -  `mouse_left_moved(...)`
  -  `mouse_left_moved(...)`
  -  `mouse_left_released(...)`
- Ctrl + mouse click moves fine, e.g. `mouse_left_pressed_control`
- Alt + mouse click moves fine, e.g. `mouse_left_pressed_meta`
- Ctrl + Alt + mouse clicks works, too
- Shift + mouse click selects words and is not detected

Basically the same behaviour with xterm, except:
- Things work fine even before the first mouse click
- Moves already report pressed modifiers (with urxvt I always get the modifiers active at the time of the last mouse click)
- Ctrl + click is caught by xterm and not reported

The `dropdown` example works fine.

I'm out of other ideas to try.

ArthurSonzogni:
Thank you very much!

psychon:
Thank *you* for fixing this so quickly

Issue #798, opened 2023-12-30
Checkbox does not `TakeFocus()` when clicked


jason-e:
When clicking a Checkbox component, its state changes as expected, but it does not call `TakeFocus()`. This means some other component will still have focus and will receive keyboard input. It also means that the only way to toggle a Checkbox with the space or enter keys is by first navigating to it with the keyboard.

ArthurSonzogni:
Thanks for reporting this! Would you like to submit a PR fixing this?

jason-e:
Done!

ArthurSonzogni:
Thanks!

PR #810, opened 2024-01-16
Make Checkbox take focus when clicked

jason-e:
Interacting with a component is a strong indicator of focus, so clicking a checkbox should cause it to take focus. This also allows keyboard input to toggle a checkbox after clicking it. (#798)


ArthurSonzogni:
Thanks!

PR #819, opened 2024-02-22
Feature: adding exit button in the homescreen example

faizan171997:
- Currently there is no clean way to exit an example (we need to press ctrl + c to do so)

- Added a exit button in the homescreen example to propose a cleaner way to exit the execution.

image


ArthurSonzogni (approved):
Thanks!

PR #827, opened 2024-03-15
Feature: add gauge_file_copy example

jopadan:
Based on `gauge` example and [cpv](https://github.com/nachoparker/cpv) allows copying files in block_size/step size blocks displaying gauge progress.
Defaults to `CPVG_BLOCK_SIZE = (2 << 12) * (2 << 5);` and overwrite if target exists by default without asking.


ArthurSonzogni:
Thanks! This looks nice.

I don't think adding this example would be helpful for FTXUI users. Indeed, this is not so much focused on FTXUI API, but more on a practical use case.
Maybe this can be its own repository linked from the FTXUI README project?

PR #831, opened 2024-03-25
Problem with setting the cursor position on the right screen edge when drawing.

jglanycon:
When moving the cursor back to its original location, a problem arises when cursor placed in the right edge column, where an off by one error occur. This pull request will resolve this problem.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Ignore all the files, except the ones we expect.
 # See https://jasonstitt.com/gitignore-whitelisting-patterns
 *
+!*/
 
 # Allowed top-level files:
 !.clang-format
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 # Ignore all the files, except the ones we expect.
 # See https://jasonstitt.com/gitignore-whitelisting-patterns
 *
-!*/
 
 # Allowed top-level files:
 !.clang-format
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ current (development)
 - Bugfix: `Input` `onchange` was not called on backspace or delete key.
   Fixed by @chrysante in chrysante in PR #776.
 - Bugfix: Propertly restore cursor shape on exit. See #792.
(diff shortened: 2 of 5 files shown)


ArthurSonzogni:
Thanks!

This is a subtle thing. Could you explain what is your problem?

I tried your patch. On `kitty` terminal, the `./examples/component/ftxui_example_input` shows an input with a cursor with a off-by-one error. It appears shifted on the right. It works correctly with `gnome-terminal`.

Without your patch, it works correctly with both terminals.

jglanycon:
If you take the same example “./examples/component/ftxui_example_input” and remove the border. Write then something into the input field until you reaching the second last character.

The cursor will then not move to the last position of the screen, due to that the cursor move escape sequence will convert the amount parameter from 0 to 1 according to the semi terminal standard.

The problem with the Kitty terminal application is related to a bug in the application itself. Because, if I switch the order of how the cursor is moved, it then works. This shouldn’t be the case, because the two escape sequences are unrelated according to the standard.

I have made a update, where I have switch the order and it should now work in all the different terminal applications :-).

jglanycon:
Link to a web page which describes the different escape sequences in details:
https://terminalguide.namepad.de

ArthurSonzogni:
You're absolutely right about the bugs in both Kitty and the last cursor position behavior in FTXUI:
- Despite the Kitty bug, we still want to support all of its users (like me).
- The FTXUI bug is not very annoying. We should fix it if it doesn't involve tradeoff.

I appreciate you sharing your latest patchset. I noticed that in both Kitty and gnome-terminal, the `ftxui_example_gallery` example exhibits an issue where the cursor on the Input is misplaced.

Interestingly, I encountered a similar challenge about a year ago. The current solution was decided as the "best" I found.

While I understand your proposed patch's intent, I agree that in its current form, it might introduce usability issues for users, so we should find a better solution.

jglanycon:
I have now reintroduced the extra move back on the x-axis, if we are not at the edge of the screen. This seems to make all the different inputs and screen types to work :-).

ArthurSonzogni:
Thanks for your contribution!

PR #832, opened 2024-03-26
Enable raw keyboard input

jglanycon:
In order for applications to receive all keyboard inputs, including the Ctrl-C and Ctrl-Z, the raw input mode has been enabled.

As result the SIGINT will no longer be used, instead the keyboard Ctrl-C event is used for exiting the framework, but only if no components has made use of it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ current (development)
 ---------------------
 
 ### Component
+- Feature: Add support for raw input. Allowing more keys to be detected.
+- Feature: Add `Mouse::WeelLeft` and `Mouse::WeelRight` events on supported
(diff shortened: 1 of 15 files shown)


ArthurSonzogni:
Excellent! Thank you very much!

jglanycon:
I have reintroduced the handling of SIGINT signal, even though I am not sure it will be used anymore. All tests now passed and you should be able to merge it into the main branch :-)

ArthurSonzogni:
After this patch, FTXUI is not able to react accordingly to (CTRL-Z aka SIGTSTP). I will need to find a solution.

ArthurSonzogni:
Merged!

PR #842, opened 2024-04-16
Fix ResizableSplit handling keyboard navigation incorrectly

na-trium-144:
example:
```cpp
auto focusableText = [] {
  return Renderer([](bool focused) {
    if (focused) {
      return text("focused!") | focus;
    } else {
      return text("not focused");
    }
  });
};
int top_size = 1;
screen.Loop(ResizableSplitTop(focusableText(), focusableText(), &top_size));
```

output:
```
focused!
--------------
not focused
```

Expected behavior: The text changes when I press up/down keys.
Actual behavior: The text changes when I press left/right keys.

I think this is because ResizableSplit internally uses Container::Horizontal regardless of its direction option, so I fixed it to use correct container.


ArthurSonzogni:
Excellent!

PR #845, opened 2024-04-18
clarified README and added tip on linking

cole-io:
clarified some sentences, changed "external package" section to "utilization", added a tip on linking for a trouble I ran into


ArthurSonzogni (approved):
Thanks!

Issue #846, opened 2024-04-27
Improper exiting with signals


ClancyWalters:
```c++
#include <ftxui/screen/screen.hpp>
#include <ftxui/component/screen_interactive.hpp>
#include <ftxui/component/component.hpp>
#include <ftxui/dom/elements.hpp>

int main(int argc, char* argv[]) {

    auto screenInteractive = ftxui::ScreenInteractive::Fullscreen();
    
    // Render a screen of nothing as an example
    auto window = ftxui::Renderer([&] {
        auto elements = ftxui::Elements {};
        for (int i = 0; i < screenInteractive.dimy(); i++) {
            elements.push_back(ftxui::text(""));
        }
        return ftxui::vbox(elements);
    });

    // On enter, gracefully exit
    window |= ftxui::CatchEvent([&](ftxui::Event event) {
        if (event == ftxui::Event::Return) {
            screenInteractive.Exit();
            return true;
        }
        return false;
    });

    std::cout << "This output should always be visable after exiting the program, missing with SIGINT" << std::endl;
    std::cout << "This might be seen and isn't cross platform";

    screenInteractive.Loop(window);
    return 0;
}
```

There seems to be a couple of (probably) internal bugs in FTXUI when handling this code. 

Firstly exiting with `SIGINT` in this case causes the screen to overwritten after being restored.
> This is due to `PostMain` being called by `~Loop()` when exiting with `screenInteractive.Exit()` whereas when exiting with `SIGINT`, `PostMain` is called by `ExecuteSignalHandlers()`. 

I suggest changing this:
```c++
void ScreenInteractive::Signal(int signal) {
  if (signal == SIGABRT) {
    OnExit();
    return;
  }
```
to this to delay execution of `PostMain()`:
```c++
void ScreenInteractive::Signal(int signal) {
  if (signal == SIGABRT) {
    Exit();
    return;
  }
```

Secondly on `bash` & `visual studio developer command prompt` the second `std::cout` is visible however on PowerShell it's overwritten. I don't really care about this issue but I guess it's worth mentioning.

PR #847, opened 2024-04-27
improving sigint handling

ClancyWalters:
PR for proposed change from #846


ArthurSonzogni:
Thanks!

I believe the correct patch was:
https://github.com/ArthurSonzogni/FTXUI/pull/848

However, I merged yours, because I like it better.

PR #849, opened 2024-04-28
Allow user to specify window element border.

hexmarsh:
It'd be neat to allow the window element to have different border styles.


ArthurSonzogni:
Thanks!

PR #867, opened 2024-05-24
Solve issues with atomic copy

felrock:
Found a small issue when playing around with making bazel build file for this repo, I could make a pull for that as well if you wish.


ArthurSonzogni:
Thanks! What is the difference?

>  I could make a pull for that as well if you wish.

Yes, please do ;-)

felrock:
The atomic class did not have a copy-constructor with I think it was GCC(C++17), but it might have been another version now that I'm remembering.

PR #868, opened 2024-05-27
fix a small bug in button example

sAkuraOfficial:
Hi!I find a small bug in the exapmle (button.cpp).
image


ArthurSonzogni:
Thanks you very much!

PR #890, opened 2024-06-25
Add missing include guard for screen/pixel.hpp

LiAuTraver:
I'm not sure whether it's intended or just oversighted, but the header `pixel.hpp` does not have include guard, for it's only included in `image.hpp` so there hadn't had any issue while using. I just added include guard:
```cpp
#ifndef FTXUI_SCREEN_PIXEL_HPP
#define FTXUI_SCREEN_PIXEL_HPP
...
#endif // FTXUI_SCREEN_PIXEL_HPP
```


ArthurSonzogni:
Thank you very much!

PR #894, opened 2024-07-02
add pixelate func and example

josias2oo3:
I added a function that takes an image and prints it onto the Terminal

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ project(ftxui
 )
 
 option(FTXUI_QUIET "Set to ON for FTXUI to be quiet" OFF)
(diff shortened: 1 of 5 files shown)


ArthurSonzogni:
Hello!

I appreciate you putting together this example – it's a great showcase of FTXUI's capabilities.

I'm weighing the value of having a visual example against the potential inconvenience of adding the OpenCV dependency. It's a trade-off, but I'm leaning towards keeping FTXUI as lightweight as possible for the majority of users.

Would you be willing to host the example in a separate repository and link to it from the README? This strikes a good balance, as it allows interested users to explore this without imposing extra dependencies on everyone else.

ArthurSonzogni:
Hello, do you have any updates?

ArthurSonzogni:
Closing, because there was no update.

PR #900, opened 2024-07-18
clang(windows) getenv deprecation warn fix

alexv-ds:
getenv in clang(windows) is deprecated

based on [spdlog](https://github.com/gabime/spdlog/blob/885b5473e291833b148eeac3b7ce227e582cd88b/include/spdlog/details/os-inl.h#L566)

### DIFF
--- a/src/ftxui/screen/terminal.cpp
+++ b/src/ftxui/screen/terminal.cpp
@@ -48,12 +48,25 @@ Dimensions& FallbackSize() {
   return g_fallback_size;
 }
 
-const char* Safe(const char* c) {
-  return (c != nullptr) ? c : "";
-}
-
 bool Contains(const std::string& s, const char* key) {
   return s.find(key) != std::string::npos;
+}
+
+// https://github.com/gabime/spdlog/blob/885b5473e291833b148eeac3b7ce227e582cd88b/include/spdlog/details/os-inl.h#L566
+std::string getenv_safe(const char *field) {
+#if defined(_MSC_VER)
+#if defined(__cplusplus_winrt)
+  return std::string{};  // not supported under uwp
+#else
+  size_t len = 0;
+  char buf[1024];
+  bool ok = ::getenv_s(&len, buf, sizeof(buf), field) == 0;
+  return ok ? buf : std::string{};
+#endif
+#else  // revert to getenv
+  char *buf = ::getenv(field); // NOLINT(*-mt-unsafe)
+  return buf ? buf : std::string{};
+#endif
 }
 
 Terminal::Color ComputeColorSupport() {
@@ -61,12 +74,12 @@ Terminal::Color ComputeColorSupport() {
   return Terminal::Color::TrueColor;
 #endif
 
-  std::string COLORTERM = Safe(std::getenv("COLORTERM"));  // NOLINT
+  std::string COLORTERM = getenv_safe("COLORTERM");
   if (Contains(COLORTERM, "24bit") || Contains(COLORTERM, "truecolor")) {
     return Terminal::Color::TrueColor;
   }
 
-  std::string TERM = Safe(std::getenv("TERM"));  // NOLINT
+  std::string TERM = getenv_safe("TERM");
   if (Contains(COLORTERM, "256") || Contains(TERM, "256")) {
     return Terminal::Color::Palette256;
   }


src/ftxui/screen/terminal.cpp
```diff
@@ -48,25 +48,38 @@ Dimensions& FallbackSize() {
...
 
+// https://github.com/gabime/spdlog/blob/885b5473e291833b148eeac3b7ce227e582cd88b/include/spdlog/details/os-inl.h#L566
+std::string getenv_safe(const char *field) {
+#if defined(_MSC_VER)
+#if defined(__cplusplus_winrt)
+  return std::string{};  // not supported under uwp
+#else
+  size_t len = 0;
+  char buf[1024];
+  bool ok = ::getenv_s(&len, buf, sizeof(buf), field) == 0;
```
ArthurSonzogni, line 63: `getenv_s` is a C11 function. Is it guaranteed to be defined in C++11?

src/ftxui/screen/terminal.cpp
```diff
@@ -48,25 +48,38 @@ Dimensions& FallbackSize() {
...
-
 bool Contains(const std::string& s, const char* key) {
   return s.find(key) != std::string::npos;
 }
 
+// https://github.com/gabime/spdlog/blob/885b5473e291833b148eeac3b7ce227e582cd88b/include/spdlog/details/os-inl.h#L566
+std::string getenv_safe(const char *field) {
+#if defined(_MSC_VER)
+#if defined(__cplusplus_winrt)
+  return std::string{};  // not supported under uwp
```
ArthurSonzogni, line 59: Does it means FTXUI won't work under UWP? This sounds problematic.

src/ftxui/screen/terminal.cpp
```diff
@@ -48,25 +48,38 @@ Dimensions& FallbackSize() {
...
+#else
+  size_t len = 0;
+  char buf[1024];
+  bool ok = ::getenv_s(&len, buf, sizeof(buf), field) == 0;
+  return ok ? buf : std::string{};
+#endif
+#else  // revert to getenv
+  char *buf = ::getenv(field); // NOLINT(*-mt-unsafe)
```
ArthurSonzogni, line 67: I believe the previous implementation was meant to fix this `*-mt-unsafe` warning. We probably want to revert it back (e.g. do not store the output in a non temporary object)

src/ftxui/screen/terminal.cpp
```diff
@@ -48,25 +48,38 @@ Dimensions& FallbackSize() {
...
 
-const char* Safe(const char* c) {
-  return (c != nullptr) ? c : "";
-}
-
 bool Contains(const std::string& s, const char* key) {
   return s.find(key) != std::string::npos;
 }
 
+// https://github.com/gabime/spdlog/blob/885b5473e291833b148eeac3b7ce227e582cd88b/include/spdlog/details/os-inl.h#L566
```
ArthurSonzogni, line 55: I don't see any warning on my bots:
https://github.com/ArthurSonzogni/FTXUI/actions/runs/9695462958/job/26755276508

What configs should I add to reproduce?


Are you using the build config the FTXUI project define (e.g. are you using CMake at the end)

ArthurSonzogni:
Thanks!

src/ftxui/screen/terminal.cpp
alexv-ds: Oh, I'm sorry. This is not msvc, but clang for windows.

```bash
cmake -Bbuild-clang -DCMAKE_CXX_COMPILER=clang++.exe -DCMAKE_C_COMPILER=clang.exe -GNinja
```

src/ftxui/screen/terminal.cpp
alexv-ds: To be honest, I haven't tested this under UWP, but I can assume that UWP applications don't have access to environment variables.

src/ftxui/screen/terminal.cpp
alexv-ds: I don't know.😅

src/ftxui/screen/terminal.cpp
alexv-ds: @ArthurSonzogni see workflow run on my fork

[https://github.com/alexv-ds/FTXUI/actions/runs/10021288139/job/27699782082#step:6:72](https://github.com/alexv-ds/FTXUI/actions/runs/10021288139/job/27699782082#step:6:72)

ArthurSonzogni:
Thanks @alexv-ds for the report and the patch, and sorry for the very long wait!

I'm going to follow a slightly different approach in #1350:

- A single `util::GetEnv()` helper keeps using `std::getenv`, and silences the deprecation warning locally around that one call with `#pragma`s (`#pragma clang diagnostic` for clang, `#pragma warning(disable : 4996)` for MSVC).
- Every call site in `terminal.cpp` and `app.cpp` goes through it.

Compared to `getenv_s`, this keeps the behavior identical: no fixed-size buffer, and no special case returning an empty value under UWP. It also covers the new `getenv` calls added since this PR.

I verified it by simulating the Microsoft CRT deprecation with clang on Linux, but I can't test a real clang build on Windows. If you still use that setup, could you confirm the warning is gone with #1350?

Closing in favor of #1350.

PR #901, opened 2024-07-18
add clion paths to gitignore

alexv-ds:
Clion creates some folders that would be nice to add to gitignore

`.idea` - local config dir
`cmake-build-*` - default build dirs for cmake (cmake-build-Debug/cmake-build-Release/etc)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,8 @@ out/
 # tools directory:
 !tools/**/*.sh
 !tools/**/*.cpp
+
+# clion config dir
+.idea/
+# clion build dirs
+cmake-build-*


ArthurSonzogni:
I'd like to discuss a couple of points regarding the proposed change:

1. **Existing .gitignore**: As you might know, FTXUI already [ignores all files](https://github.com/ArthurSonzogni/FTXUI/blob/307e4eb4b3e0f245b85312e0bd028d748f94b740/.gitignore#L1-L4) by default in its .gitignore file: Given this, I don't see a benefit with this specific addition. Did you encounter any issues where excluding this directory was necessary?

2. **Maintaining .gitignore**: While .gitignore files can be helpful, managing them for specific developer workflows can become cumbersome and lead to frequent pull requests. Ideally, FTXUI shouldn't rely on individual preferences.  A better approach might be for developers to leverage their own global `.gitignore` files for environment-specific needs.

What do you think? Perhaps there's another aspect to consider that I might have missed.

Thanks!

alexv-ds:
No, you haven't missed anything, this is nothing more than my personal preference. The global .gitignore argument sounds good.

PR #938, opened 2024-10-18
Add `SliderWithCallback` component

mikolajlubiak:
I wanted to do some actions in real time when the slider value changes, not only update the reference value, so I made a little `SliderWithCallback` component

### DIFF
--- a/src/ftxui/component/slider.cpp
+++ b/src/ftxui/component/slider.cpp
@@ -256,6 +256,185 @@ class SliderWithLabel : public ComponentBase {
   Box box_;
   bool mouse_hover_ = false;
 };
+
+template <class T>
+class SliderWithCallback : public ComponentBase {
+public:
+  explicit SliderWithCallback(SliderWithCallbackOption<T> options)
+      : callback_(options.callback), min_(options.min), max_(options.max),
+        increment_(options.increment), options_(options) {
+    SetValue(options.value());
+  }
+
+  Element Render() override {
+    auto gauge_color = Focused() ? color(options_.color_active)
+                                 : color(options_.color_inactive);
+    const float percent = float(value_() - min_()) / float(max_() - min_());
+    return gaugeDirection(percent, options_.direction) |
+           flexDirection(options_.direction) | reflect(gauge_box_) |
+           gauge_color;
+  }
+
+  void OnLeft() {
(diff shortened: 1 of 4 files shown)


mikolajlubiak:
Example usage: https://github.com/mikolajlubiak/memory/

ArthurSonzogni:
Thanks! This looks like a good thing to have. 

Could you please extend the current `SliderOption` instead? This would avoid duplicating the current code.

mikolajlubiak:
@ArthurSonzogni friendly ping

ArthurSonzogni:
Thanks! LGTM.

- I renamed `callback` into `on_change`.
- I added tests.
- I fixed bugs where `on_change` was called multiple time per change.
- I added the change to the CHANGELOG.

If anything looks wrong to you, please let me know, and we will maek the appropriate change.

src/ftxui/component/slider.cpp
```diff
@@ -127,8 +123,11 @@ class SliderBase : public ComponentBase {
       OnUp();
     }
 
-    SetValue(util::clamp(value_(), min_(), max_()));
-    if (old_value != value_()) {
+    this->value() = std::max(this->min(), std::min(this->max(), this->value()));
```
mikolajlubiak, line 126: Why use this instead of std::clamp?

mikolajlubiak:
To get to the contributors list I would have to make more contributions?

mikolajlubiak:
Also would it be cool to add my memory game to the projects using FTXUI list or is it for more advanced projects?

ArthurSonzogni:
> To get to the contributors list I would have to make more contributions, right?

You are in the contributor list:
image

It might take a few hours for the displayed image to be invalidated and the new version to appear.

> Also would it be cool to add my [memory game](https://github.com/mikolajlubiak/memory/) or [media (video/gif/image) into colored, animated or static, ASCII art converter](https://github.com/mikolajlubiak/terminal_animation) to the projects using FTXUI list or is it for more advanced projects?

It would be cool. Do you want to submit another PR adding it?

mikolajlubiak:
Thanks, I made the PR: #946

`std::max(this->min(), std::min(this->max(), this->value()));`
Out of curiosity, why use this instead of `std::clamp`?

PR #944, opened 2024-10-29
Add 2048-cpp to projects using FTXUI

Chessom:
I implement the 2048 game using FTXUI. This library is really powerful and amazing!


ArthurSonzogni:
Awesome! I tried it, and I love the smooth transition, and the "Automatic Play Switch".

Issue #951, opened 2024-11-10
Resize ScreenInteractive::TerminalOutput() on SIGWINCH issue


dangooddd:
Hi!
I got problem with terminal resizing and ScreenInteractive::TerminalOutput component. This issue appears not only in tmux, but also in plain terminal (tested on Kitty, Alacritty, Foot and Konsole). Redrawing works well when fullscreen.

How can I solve this? Compiled with clang 19, version of ftxui is 5.0.0

I am new to the TUI thing and this community, so sorry if this issue well known, I can't find anything about this.

[redrawing_issue.webm](https://github.com/user-attachments/assets/e4d26fc6-293d-4086-be93-94c9341840f0)

dangooddd:
Code is almost identical with menu example from repo
```cpp
#include "embf/directory.hpp"
#include "ftxui/component/component.hpp"
#include "ftxui/component/component_options.hpp"
#include "ftxui/component/screen_interactive.hpp"
#include "ftxui/dom/elements.hpp"
#include "ftxui/screen/screen.hpp"
#include <ftxui/dom/elements.hpp>
#include <iostream>

int main(int argc, char *argv[]) {
    using namespace ftxui;
    auto screen = ScreenInteractive::TerminalOutput();

    std::vector<std::string> entries = {
        "entry 1",
        "entry 2",
        "entry 3",
    };
    int selected = 0;

    MenuOption option;
    option.on_enter = screen.ExitLoopClosure();
    auto menu = Menu(&entries, &selected, option) | borderLight;

    screen.Loop(menu);

    std::cout << "Selected element = " << selected << std::endl;

    return 0;
}
```

ArthurSonzogni:
Hello,
Line wrapping are implemented by the terminal emulator. I don't think I know any solution against this.

If you used `ScreenInteractive::Fullscreen` we would draw over the whole area, on an alternate screen, which would avoid seeing all the garbage.

If you know about any other solution, I would happily implement it.

dangooddd:
I see that fzf does this, but only when resize left/right. I guess that after sigwinch it redraws window at top of terminal, so it becomes fullscreen-like. See video below.

I think that is not a bad solution, because anyway output of terminal becomes crap, so why not to clear it? What do you think about this?
[fzf_resizing.webm](https://github.com/user-attachments/assets/62198d74-66b3-48f0-a1c7-82bd48a2c835)

ArthurSonzogni:
[...]

PR #952, opened 2024-11-12
Clear terminal output of interactive screen if alternate screen not in use

dangooddd:
This PR is submitted to close https://github.com/ArthurSonzogni/FTXUI/issues/951.

Few moments about changes:
1) Interactive screen is always created with dimx = 0 and dimy = 0, so on first iteration of loop screen will be resized.
I add new variable validated_ that is true since second iteration.
I think we can avoid this by passing component_ to the PreMain function to set init dimx and dimy, but than we should add component to the Install function, and than to the PostMain function...
2) Some part of scrollback (no more than one terminal screen) will be cleared because of how I chose to implement this, but i do not think it is big problem

### DIFF
--- a/include/ftxui/component/screen_interactive.hpp
+++ b/include/ftxui/component/screen_interactive.hpp
@@ -121,6 +121,7 @@ class ScreenInteractive : public Screen {
   bool mouse_captured = false;
   bool previous_frame_resized_ = false;
 
+  bool validated_ = false;
   bool frame_valid_ = false;
 
   bool force_handle_ctrl_c_ = true;
--- a/src/ftxui/component/screen_interactive.cpp
+++ b/src/ftxui/component/screen_interactive.cpp
@@ -858,6 +858,14 @@ void ScreenInteractive::Draw(Component component) {
   ResetCursorPosition();
   std::cout << ResetPosition(/*clear=*/resized);
 
+  // clear terminal output if non-fullscreen interactive screen dimx decreases
+#if !defined(_WIN32)
+  if ((dimx < dimx_) && validated_ && !use_alternative_screen_) {
+    std::cout << "\033[J";
+    std::cout << "\033[H";
+  }
+#endif
+
   // Resize the screen if needed.
   if (resized) {
     dimx_ = dimx;


dangooddd:
I think it works the way fzf does this. Results:

[clear_on_resize.webm](https://github.com/user-attachments/assets/71e17760-4e42-4c07-8f16-677bfd80bad3)

include/ftxui/component/screen_interactive.hpp
```diff
@@ -121,6 +121,7 @@ class ScreenInteractive : public Screen {
   bool mouse_captured = false;
   bool previous_frame_resized_ = false;
 
+  bool validated_ = false;
```
ArthurSonzogni, line 124: Why do we need this variable?

I see it can move from false->true on the first drawing, but then it is never reset back to false.

src/ftxui/component/screen_interactive.cpp
```diff
@@ -858,6 +858,16 @@ void ScreenInteractive::Draw(Component component) {
   ResetCursorPosition();
   std::cout << ResetPosition(/*clear=*/resized);
 
+  // clear terminal output if screen dimx decreases
+  // only on primary screen
+  // only on POSIX systems (linux/macos)
+#if !defined(_WIN32)
```
ArthurSonzogni, line 864: Is there any reasons to exclude windows?

src/ftxui/component/screen_interactive.cpp
```diff
@@ -858,6 +858,16 @@ void ScreenInteractive::Draw(Component component) {
   ResetCursorPosition();
   std::cout << ResetPosition(/*clear=*/resized);
 
+  // clear terminal output if screen dimx decreases
```
ArthurSonzogni, line 861: It would be great explaining why (and not what) we do this.

What about:
```cpp
If the terminal width decrease, the terminal emulator will start wrapping lines and make the display dirty. We should clear it completely.
```

ArthurSonzogni (commented):
Excellent! Thanks!

I approved % minor comments below:

include/ftxui/component/screen_interactive.hpp
dangooddd: I figure out that we do not need this variable in case when clearing appears only after decrease of dimx, so I will remove it. Thanks

src/ftxui/component/screen_interactive.cpp
dangooddd: I was not able to test this on windows, so i decided to exclude it. As i can see, windows supports this escape sequences so i will remove that

src/ftxui/component/screen_interactive.cpp
dangooddd: Great explain, thanks

ArthurSonzogni:
Thanks you!

Issue #904, opened 2024-07-24
Changes cursor in underlying terminal


aaron-michaux:
When I run an ftxui "window", it changes my cursor to blinking block. Then, when I quit the application, my terminal's cursor remains changed.

cole-io:
This is happening to me too in Windows Terminal. It's important to say that I am using WSL.

Even when I change the cursor back to a bar using `Screen::SetCursor` before the application terminates, the cursor still appears as a block after the process exits.

cole-io:
Possibly related: https://github.com/neovim/neovim/issues/4396

PR #953, opened 2024-11-13
Add option to change cursor shape on exit of ScreenInteractive loop

dangooddd:
Currently there is no option to set cursor_reset_shape_ on ScreenInteractive class. Because of that, cursor is always block and blinks (cursor shape is always 1). 

This PR adds SetCursorReset method as option.

Will close https://github.com/ArthurSonzogni/FTXUI/issues/904

### DIFF
--- a/include/ftxui/component/screen_interactive.hpp
+++ b/include/ftxui/component/screen_interactive.hpp
@@ -39,6 +39,8 @@ class ScreenInteractive : public Screen {
   // Options. Must be called before Loop().
   void TrackMouse(bool enable = true);
 
+  void SetCursorReset(Screen::Cursor::Shape shape);
+
   // Return the currently active screen, nullptr if none.
   static ScreenInteractive* Active();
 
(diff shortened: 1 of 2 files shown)


ArthurSonzogni:
Thanks!

I don't think it works. FTXUI request the terminal to reflect what is the current cursor [here (1)](https://github.com/ArthurSonzogni/FTXUI/blob/main/src/ftxui/component/screen_interactive.cpp#L604), save it [here (2)](https://github.com/ArthurSonzogni/FTXUI/blob/main/src/ftxui/component/screen_interactive.cpp#L773), and restores it [here (3)](https://github.com/ArthurSonzogni/FTXUI/blob/main/src/ftxui/component/screen_interactive.cpp#L607)

I believe the value set by this new API would be overriden by (2), doesn't it?


Why do you need this API? Are you suggesting the current way of requesting/restoring the cursor value doesn't work?

dangooddd:
Sorry, i just had not seen https://github.com/ArthurSonzogni/FTXUI/blob/main/src/ftxui/component/screen_interactive.cpp#L604. Indeed, restoring cursor this way will be better than set this manually with New API.

Problem is that in my case ftxui always sets blinking cursor no matter what i do. I will try to debug this and describe the problem latter. Also in https://github.com/ArthurSonzogni/FTXUI/issues/904 I guess problem exactly like i have

Issue #940, opened 2024-10-21
Properly changing window title text color


ArthurSonzogni:
### Discussed in https://github.com/ArthurSonzogni/FTXUI/discussions/939

<div type='discussions-op-text'>

<sup>Originally posted by **sentientbottleofwine** October 20, 2024</sup>
This might be a very simple question and I might be just missing something obvious, I tried and couldn't resolve my issue. When I try to make a window that has a title that's supposed to be colored and apply the color to the title the whole upper border of the window changes color:
![20241020_17h22m27s_grim](https://github.com/user-attachments/assets/92377408-78a9-4531-a6b6-8356d272c895)
Here's the code:
```cpp
#include <ftxui/dom/elements.hpp>
#include <ftxui/screen/screen.hpp>

int main() {
    using namespace ftxui;
    auto render = window(text("Only text should be colored") | color(Color::Green), text("Contents")) | color(Color::White);
    auto screen = Screen::Create(
        Dimension::Full(),
        Dimension::Fit(render)
    );
    Render(screen, render);
    screen.Print();
}
```
I have no idea how to do this correctly. I greatly appreciate and thank any help.</div>

ArthurSonzogni:
Thanks!

This looks like a bug. I believe this is because the "title" is flexible by default and covers the whole line, causing the color decorator to apply to everything that is behind.

We probably want to fix this.

sentientbottleofwine:
Thanks for the response!
So I tried looking into this and from what I can see the `title_box` has it's `x_min` and `x_max` set to the `x_min` and `x_max` of the whole window, this creates a problem where a color decorator sets the color for all pixels in the `title_box` including the ones from the original border because of the previously mentioned value assigned to `title_box`. Just like you said! But if we try to decrease those values if an `hcenter` decorator has been applied. We can see that the title of the window is shifted by the amount we decreased from the center. I can't figure out how to change the color of only the title without shifting it (Also I don't know where to get the 'true' values for the `title_box` anyway). @ArthurSonzogni Let me know what do you think.

PR #961, opened 2024-11-30
Fix: Properly changing window title text color #940

Chessom:
Just change this
```cpp
title_box.x_max = box.x_max - 1;
```
to this
```cpp
title_box.x_max = std::min(box.x_max - 1, box.x_min + children_[1]->requirement().min_x);
```


ArthurSonzogni:
Excellent! Thanks!

PR #977, opened 2024-12-27
Implement Node::Select for flexbox.

ArthurSonzogni:


ArthurSonzogni:
+CC @clement-roblot FYI.

There are still a few issues. Eventually, this should fix how a flexbox is selected.

clement-roblot:
How convenient! I come back with a plan to look into this issue and you already solved it :heart: 

This commit fixes the issue I mentioned in the MR #926  :+1:

PR #978, opened 2024-12-27
Merge dom and component focus

### LINKED ISSUE #935: Last scroll position of the vertical/horizontal container being reset when it's out of focus.

horrifyingHorse:
Following the approach discussed in #970 to resolve #935


horrifyingHorse:
Greetings,

I suppose it's fixed now, although i would like to add that i had to accommodate the `SetCursor` of  `Focus::Render` in `Node::Render`, anywhere else, it kept messing up the cursor on input.

Is there any better way to do it?

ArthurSonzogni:
Excellent!
At first sight, it looks exactly as I hoped!

I am quite busy at the moment. I will take a deeper look as soon as possible. Thank you very much for making this patch!

horrifyingHorse:
There is an issue with the cursor. In `Input` component, the cursor is not positioned correctly inside the input box.

I've cross checked it with the previous commits and it seems the issue arises in [`511a294`](https://github.com/ArthurSonzogni/FTXUI/pull/978/commits/511a2941672bced636580448fc5ceeb5cf9857f6)
<hr>

`to recreate`
```cpp
#include <ftxui/component/component_options.hpp>
#include "ftxui/component/component.hpp"
#include "ftxui/component/screen_interactive.hpp"

using namespace ftxui;

int main() {
  std::string inp;
  auto input = Input(&inp, "input", InputOption::Spacious());
  auto display = Renderer(input, [&] {
    return vbox({
      hbox({
        filler(),
        vbox({
          text("Banner") | center,
        }) | size(WIDTH, EQUAL, 40) | border,
        filler()
      }),
      filler(),
      input->Render()
        | size(WIDTH, EQUAL, 20) | size(HEIGHT, EQUAL, 7) | center
    });
  });

  screen.Loop(display);
}
```

ArthurSonzogni:
I think I fixed the last issue.

@horrifyingHorse , do you believe this patch is ready to land?

horrifyingHorse:
Yes! although a minor issue, the cursor appears to be at the box max when input is empty, works fine when its not empty.
I think this addition causes the issue
```diff
if (content->empty()) {
      auto element = text(placeholder()) | xflex | frame;

      return transform_func({
                 std::move(element), hovered_, is_focused,
                 true  // placeholder
             }) |
-             reflect(box_);
+            focused | reflect(box_);
    }

```

Screenshot_20250304_212133

horrifyingHorse:
Since we updated for elements to have `focus` only when its `Focused()`, this would always equate false when not `focus`, and the issue of scroll to top when !focus persists:
```cpp
if (child->requirement().focused.enabled) {
```
I updated the condition to fix it in my latest commit.

I think the patch is ready to land :)

ArthurSonzogni:
I ended up implementing a different idea.

The component will fill "requirement.focused.component_active" to select the right children.

PR #988, opened 2025-01-16
Add inLimbo as example project

nots1dd:
In case a description is required,

inLimbo is a TUI music player that creates a very fast, efficient and cool environment for anybody to enjoy their favorite music that works across linux distributions without a hassle.


ArthurSonzogni (approved):
Beautiful project! LGTM.

Issue #1023, opened 2025-03-30
resizable split resize crashes


littleblack111:
example of the resizable split is ScreenInteractive::Fullscreen, but for ScreenInteractive::TerminalOutput, it crashes with
```
[ERROR] Fatal error: cannot create std::vector larger than max_size()
```
 after some weirdness, hard to describe.

PR #1025, opened 2025-03-31
Bugfix: Avoid crash with ResizeableSplit.

ArthurSonzogni:
Component
---------
- Bugfix: Fix a crash with ResizeableSplit. See #1023.
  - Clamp screen size to terminal size.
  - Disallow `ResizeableSplit` with negative size.

Dom
---
- Bugfix: Disallow specifying a negative size constraint. See #1023.

Bug: https://github.com/ArthurSonzogni/FTXUI/issues/1023

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,17 @@
 Changelog
 =========
 
+Development
+-----------
+### Component
+- Bugfix: Fix a crash with ResizeableSplit. See #1023.
+  - Clamp screen size to terminal size.
+  - Disallow `ResizeableSplit` with negative size.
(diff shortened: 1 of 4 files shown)


littleblack111 (approved):
LGTM

littleblack111:
> @littleblack111, does this patch looks good to you?
> 
> > that commit seems to just make TerminalOutput same as Fullscreen?
> 
> I forgot a `break` in a switch statement. Thanks for finding it.

this is enough for avoiding the crash, however, the UI is still off, as it seems to only resize the container but not put it back to where it came from

PR #1028, opened 2025-04-03
feat(resizablesplit): min size

littleblack111:


ArthurSonzogni:
Thank you for the PR!

To keep the `ResizableSplitXxx` interfaces clean, I'd prefer to avoid adding new parameters directly.  We should leverage the `ResizableSplitOption` for this type of customization.

It appears the functionality you're aiming for can be achieved using `split |= size(HEIGHT, GREATER_THAN, 10)`.  Would this approach address your needs?  I'd like to prioritize combining existing decorators over introducing new parameters, if possible.

littleblack111:
> Thank you for the PR!
> 
> To keep the `ResizableSplitXxx` interfaces clean, I'd prefer to avoid adding new parameters directly. We should leverage the `ResizableSplitOption` for this type of customization.
> 
> It appears the functionality you're aiming for can be achieved using `split |= size(HEIGHT, GREATER_THAN, 10)`. Would this approach address your needs? I'd like to prioritize combining existing decorators over introducing new parameters, if possible.

thanks for pointing it out. it was not very obvious on the wiki, thanks

littleblack111:
however, the method you provided doesn't seem to work:
> It appears the functionality you're aiming for can be achieved using split |= size(HEIGHT, GREATER_THAN, 10). Would this approach address your needs? I'd like to prioritize combining existing decorators over introducing new parameters, if possible.

```
	container = logComponent |= ftxui::Scroller;
	container = ftxui::ResizableSplitBottom(inputComponent, container, &bottomSize);
	container |= ftxui::size(ftxui::HEIGHT, ftxui::GREATER_THAN, 10);
```

littleblack111:
> however, the method you provided doesn't seem to work:
> 
> > It appears the functionality you're aiming for can be achieved using split |= size(HEIGHT, GREATER_THAN, 10). Would this approach address your needs? I'd like to prioritize combining existing decorators over introducing new parameters, if possible.
> 
> ```
> 	container = logComponent |= ftxui::Scroller;
> 	container = ftxui::ResizableSplitBottom(inputComponent, container, &bottomSize);
> 	container |= ftxui::size(ftxui::HEIGHT, ftxui::GREATER_THAN, 10);
> ```

and actually, this broke [scroller](https://github.com/ArthurSonzogni/git-tui/blob/master/src/scroller.cpp)

Issue #1032, opened 2025-04-22
bazel build file?


kcc:
Please consider maintaining a bazel build file - will make it easier to import FTXUI into bazel-based projects.

ArthurSonzogni:
Thanks @kcc!

I am open to it.
I remember someone worked on it:
https://github.com/ArthurSonzogni/FTXUI/discussions/800
but this was not pursued.

Out of curiosity, do you have any project in mind where you want to use FTXUI?

I am going to add support, please expect 2-3 weeks. Feel free to let me know if you want to help (code review, and/or a PR)

kcc:
I am investigating my options for a TUI library that could be used to implement a specialized log viewer (multiple text view panes, navigation, highlighting, etc). 
Would be happy to help with a code review - thanks!

ArthurSonzogni:
Thanks!

@kcc, here is the PR:
https://github.com/ArthurSonzogni/FTXUI/pull/1033

> I am investigating my options for a TUI library that could be used to implement a specialized log viewer (multiple text view panes, navigation, highlighting, etc).

Cool. Please don't hesite to let me know if there is something you struggle to implement with FTXUI. Happy to give you some examples.

kcc:
Thank you! Using FTXUI in a bazel-based project is now zero-effort.

PR #1033, opened 2025-04-24
Support Bazel build system

ArthurSonzogni:
Add support for the Bazel build system, and define a worflow to publish every new version to the to Bazel Central repository.

Bug:https://github.com/ArthurSonzogni/FTXUI/issues/1032
Fixed:https://github.com/ArthurSonzogni/FTXUI/issues/1032


ArthurSonzogni:
@kcc this is ready \o/

Please let me know if you have any comments?

ArthurSonzogni (commented):
LGTM

BUILD.bazel
```diff
@@ -0,0 +1,251 @@
...
+load(":bazel/ftxui.bzl", "windows_copts")
+load(":bazel/ftxui.bzl", "pthread_linkopts")
+
+package(default_visibility = ["//visibility:public"])
+
+# A meta target that depends on all the ftxui sub modules.
+alias(
+    name = "ftxui",
+    # Note that :component depends on :dom, which depends on :screen. Bazel
+    # doesn't really support "public" and "private" dependencies. They are all
```
robinlinden, line 25: You can have public and private dependencies in Bazel by using `implementation_deps` for the private ones. :P

robinlinden (commented):
The way transitive dependencies are used here only really works if the `layering_check`-feature in Bazel is disabled. Most of the cases are pretty easy to fix by just adding the transitive dependencies, but enabling that feature also blocks things like using things from `//:screen`'s srcs as includes in `//:component`:
```
src/ftxui/component/screen_interactive.cpp:37:10: error: use of private header from outside its module: 'ftxui/screen/util.hpp' [-Wprivate-header]
   37 | #include "ftxui/screen/util.hpp"                      // for util::clamp
      |          ^
1 error generated.
```

There's more info about Bazel's header inclusion checking here: https://bazel.build/reference/be/c-cpp#hdrs

I've updated to v6.1.8 and dropped our Bazel port of ftxui, but disabled the `layering_check` only for ftxui (using patch_cmds) for now: https://github.com/robinlinden/hastur/pull/1255/commits/d16bec7da86c6ba10b745be4e3dc5a3f607fa745

Thank you for the official Bazel support. :)

BUILD.bazel
ArthurSonzogni: Thanks for letting me know!

PR #1043, opened 2025-05-04
Bazel: general improvements.

ArthurSonzogni:
Bazel: general improvements.

Improve the Bazel build. Attempt to fix previous errors recorded while
trying to publish ftxui in the Bazel Central Registry:
- https://github.com/bazelbuild/bazel-central-registry/pull/4485
- https://buildkite.com/bazel/bcr-presubmit/builds/13601#01968b61-f5b2-4d16-94d0-c87a03a1a23b

Test against "recent" platforms
-------------------------------

Previously, I got the error:
```
gcc: error: unrecognized command line option '-std-c++20'; did you mean '-std-c++2a'?
```
This was due to using old distribution like ubuntu 2004. Test against
newer platforms only to avoid GCC version<-9.x.y

Downgrade gtest version.
------------------------

I suspect this caused the Bazel Central Registry error:
```
file:///workdir/modules/googletest/1.15.2/MODULE.bazel:68:20: name 'use_repo_rule' is not defined
```
I hoped using a lower version will fix the issue.

Tag gtest as dev_dependency
---------------------------

Presumably, this should avoid dependants to fetch it?

Enable --features-layering_check
--------------------------------

Aka clang `-Wprivate-header`. Fix the encountered errors.

Use clang in the CI
-------------------

The CI was defining clang/gcc in the matrix, but was not using it. Fix
the bug.


ArthurSonzogni:
Hello @kcc and @robinlinden,

Would you have any additional comments / suggestion about FTXUI bazel build files? [^1]

[^1]: This is my first use of Bazel.

tools/test_bazel.sh
```diff
@@ -0,0 +1,27 @@
+# This script tests the project with different versions of Bazel and compilers
+# locally. This avoids waiting on the CI to run the tests.
+
+# Version
+# -------
+# - Version >= 7 are supported
+# - Version <= 6 fail with the error:
```
robinlinden, line 7: I wonder if passing `--enable_bzlmod` would fix this issue. iirc bzlmod wasn't defaulted to on until Bazel 7..

.github/workflows/build.yaml
```diff
@@ -40,10 +40,14 @@ jobs:
         uses: actions/checkout@v3
 
       - name: "Build with Bazel"
+        env:
+          CC: ${{ matrix.compiler }}
```
robinlinden, line 44: I think I've had situation where I had to set both CC and CXX for Bazel to actually use the compilers I specified and not default to gcc, but not 100% certain about this.

.bcr/presubmit.yml
```diff
@@ -1,11 +1,16 @@
 matrix:
+  # Note that gcc 9 and earlier aren't supported due to using --std=c++2a
+  # instead of --std=c++20
   platform:
-  - centos7
-  - debian10
-  - ubuntu2004
+  - debian11
+  - ubuntu2024
```
robinlinden, line 6: This should probably be 2404 instead of 2024? 2204 might also work since that has a relatively recent gcc.

MODULE.bazel
```diff
@@ -1,9 +1,13 @@
...
-# Build deps.
+# Build dependencies.
 bazel_dep(name = "rules_cc", version = "0.1.1")
-bazel_dep(name = "platforms", version = "0.0.11")
+bazel_dep(name = "platforms", version = "0.0.10")
 
-# Test deps.
-bazel_dep(name = "googletest", version = "1.16.0.bcr.1")
+# Test dependencies.
+bazel_dep(name = "googletest", version = "1.14.0.bcr.1", dev_dependency = True)
```
robinlinden, line 13: There was some discussion in bazelbuild/bazel about test-only deps as dev_dependencies: https://github.com/bazelbuild/bazel/issues/22187
iirc there's some ongoing work on optional dependencies, but I don't think that's in a state where it's usable yet.

robinlinden (commented):
Looks great! Should allow me to drop the layering_check patch I have. :)

tools/test_bazel.sh
ArthurSonzogni: It fixes the issue! Thanks!

Unfortunately, I still make it work on Bazel 6.0.0. I get:
```
external/googletest~1.14.0.bcr.1/googletest/include/gtest/internal/gtest-port.h:279:2: error: #error C++ versions less than C++14 are not supported.
```

Unsure why. Probably a problem from `googletest`.
```
ERROR: /home/arthursonzogni/.cache/bazel/_bazel_arthursonzogni/fd1242e4c381a003bc1e116d9c36c824/external/googletest~1.14.0.bcr.1/BUILD.bazel:80:11: Compiling googlemock/src/gmock-cardinalities.cc failed: (Exit 1): gcc failed: error executing command (from target @googletest~1.14.0.bcr.1//:gtest) /usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer '-std=c++0x' -MD -MF ... (remaining 34 arguments skipped)
```

.github/workflows/build.yaml
ArthurSonzogni: Thanks! Let's try to specify both.

.bcr/presubmit.yml
ArthurSonzogni: Thanks! That's a great catch!

ArthurSonzogni: I meant `ubuntu2204`

tools/test_bazel.sh
robinlinden: I guess something's defaulting to C++14, either Bazel or an older compiler. You could try adding `build --cxxopt=-std=c++17`* to your `.bazelrc`, should fix it. Note that when people depend on FTXUI using Bazel, your `.bazelrc` will be ignored, and only the one in the top-level repository will be used.

*or this if you want things to be happy on Windows as well since they use `:` instead of `=` in their std flag:
```
build --enable_platform_specific_config
build:linux --cxxopt=-std=c++17
build:macos --cxxopt=-std=c++17
build:windows --cxxopt=-std:c++17
```

ArthurSonzogni: Thanks! I will use the `.bazelrc` and abandon specifying the standard per target. This will be on dependants to do it.

Too bad Bazel doesn't support specifying the C++ standard for a target like CMake does.

PR #1061, opened 2025-06-10
added two bash scripts for easy compile

npc-gnu:
# 1. g++ftxui
## Pros: It makes compiling a lot easier. You dont prepare a .txt file. You dont use desillion -I flags. You just write:
```bash
g++ftxui -o out source_code.cpp
```
## Cons: It is bloated. It ads all of `libftxui-component.a, libftxui-dom.a, libftxui-screen.a`
# 2. install
It installs g++ftxui script.
```bash
chmod +x install && sudo ./install
```
scripts

(NOTE: I tried to change wrong projects license. Please don't see license requests. Sorry.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ Next
   ```
   Thanks @mikomikotaishi for PR #1015.
 
+### Component
+- Fix ScreenInteractive::FixedSize screen stomps on the preceding terminal
+  output. Thanks @zozowell in #1064.
+
 
 6.1.9 (2025-05-07)
 ------------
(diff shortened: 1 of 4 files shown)


ArthurSonzogni:
Thanks!

I thought I previously answered this PR, but it looks like I haven't. Sorry about this.

I remember I don't believe this would be a good idea to merge this PR. There are not good ways to build FTXUI outside of cmake/bazel for now. The proposed change looks incorrect at first glance, as it is missing a couple of compiler options.

I created new docs in:
https://arthursonzogni.github.io/FTXUI/installation.html

I wonder if we shouldn't add some text explaining how users could build/use FTXUI without depending on a build system. Unsure what we would say...

npc-gnu:
Oh okay then. 
And no problem Strawberry bro about latency.
"The proposed change looks incorrect at first glance, as it is missing a couple of compiler options."
Would you please tell what is missing?
Thank you.

PR #1069, opened 2025-06-24
Support for extracting the terminal id.

jglanycon:
I have added support for extracting the terminal id for the number of different terminal applications (Not all terminal id has been added yet)

The main idea, is to use this terminal id to enable or disable different features in the FTXUI libraries based on which terminal application been used. Second, we currently use this feature for performing colour correction between GUI terminals and the default Linux login (in text mode), where colours are bit off.

I hope this makes sense :-).

### DIFF
--- a/src/ftxui/component/button.cpp
+++ b/src/ftxui/component/button.cpp
@@ -139,7 +139,6 @@ class ButtonBase : public ComponentBase, public ButtonOption {
  private:
   bool mouse_hover_ = false;
   Box box_;
-  ButtonOption option_;
   float animation_background_ = 0;
   float animation_foreground_ = 0;
   animation::Animator animator_background_ =


ArthurSonzogni:
**Excellent!**
I wasn’t aware of the terminal API that reports the terminal ID—this is very useful! It’ll help implement dynamic fallbacks for specific terminals. For instance, we can now detect "Microsoft Windows Terminal" at runtime instead of relying on build-time flags.

I’m currently quite busy, so the patch will likely be merged later this week. That said, it looks very well written at first glance. Thanks again!

ArthurSonzogni:
Note: I am going to remove usage of threads. Then insert a startup sequence waiting synchronously for the the terminal ID. Then re-emitting all the event it received before starting the main loop. This might take me some time, but I think it's worth it having a sync API for this.

PR #1070, opened 2025-06-28
Add gridbox children to children_ 

KenReneris:
Fix so layout will iterate when there's a flex element inside of a gridbox that grows it's box size on the y axis.   (Allows need_iteration_ to work from Flexbox::SetBox() that's within the grid so that layout will re-run as the box grows)

### DIFF
--- a/src/ftxui/dom/box_helper.cpp
+++ b/src/ftxui/dom/box_helper.cpp
@@ -58,11 +58,39 @@ void ComputeShrinkHard(std::vector<Element>* elements,
 
     element.size = element.min_size + added_space;
   }
+
+}
+
+
+// Called when the size allowed is lower than the requested size, and the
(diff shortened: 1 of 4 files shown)


ArthurSonzogni (commented):
Thanks!

It looks like there is definitely a bug that this patch is fixing (adding elements as children). However I do not understand the parts about flexbox_helper.

ArthurSonzogni:
I merged the first part of this patch!

ArthurSonzogni:
Thanks @KenReneris! The main part of this patch (adding the gridbox cells to `children_`) was merged in 093d6fc40. The remaining changes (the `hack` flag in `box_helper` and removing `flex_shrink` from table cells) still need an explanation and a regression test, so I'm closing this for now. Feel free to open a new PR with a test in `gridbox_test.cpp` showing the flexbox-in-gridbox case it fixes.

PR #1084, opened 2025-07-18
Fix: Use of max() min() collides with predefined macros in Windows build

solosTec:
When compiling include/ftxui/component/component_options.hpp on windows the vc compiler gives an error:

`Error C2760	syntax error: ')' was unexpected here; expected 'expression' ... component_options.hpp	230`		

The reason for this is that max() and min() are predefined macros under Windows. The proposed solution is to remove the definition of these macros.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -2,3 +2,6 @@
 # http://clang.llvm.org/docs/ClangFormatStyleOptions.html
 BasedOnStyle: Chromium
 Standard: Cpp11
+
+InsertBraces: true
+InsertNewlineAtEOF: true
--- a/examples/component/canvas_animated.cpp
+++ b/examples/component/canvas_animated.cpp
@@ -133,8 +133,9 @@ int main() {
       float dy = 50.f;
       ys[x] = int(dy + 20 * cos(dx * 0.14) + 10 * sin(dx * 0.42));
     }
-    for (int x = 1; x < 99; x++)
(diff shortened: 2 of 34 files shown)


ArthurSonzogni:
I appreciate your contribution.

FTXUI users have the flexibility to define arbitrary macros. I don't believe it's FTXUI's responsibility to mitigate issues arising from user-defined macros, even when a common header like `windows.h` is involved.

### User Solutions

FTXUI consumers encountering this issue with `windows.h` should implement one of the following solutions:

* **Define `NOMINMAX`**: Add `#define NOMINMAX` before `#include <windows.h>`.
* **Compiler Flag**: Pass `/DNOMINMAX` as a compiler flag.

### Potential FTXUI Enhancement

A potential enhancement for FTXUI could be the introduction of a utility header, perhaps `ftxui/util/warn_windows_macro.h`. This header could detect if `min` or `max` are defined as macros and then trigger an `#error` with clear instructions on how users can resolve the conflict from their end.

What are your thoughts on this approach?

solosTec:
To define **NOMINMAX** is a good approach. The only problem I see is users who want to use these macros. I saw this problem many times when working in a windows environment. I therefore see two viable solutions:

- utility header (preferred solution)
- renaming the project-specific functions to avoid the naming conflict (possibly a breaking change)

Ok, keep up the good work

ArthurSonzogni (approved):

PR #1085, opened 2025-07-19
Add ftxui convenience/umbrella module to cmake rules to fix #1083

### LINKED ISSUE #1083: import ftxui; // fails due to missing src/ftxui/ftxui.cppm line in ftxui_modules.cmake
Not sure if there is a reason it's not listed in the modules cmake file. It was discussed in the PR #1015 but I don't see it in the actual PR commit, and the (script edited) code build examples don't use `import ftxui;` though the docs say it.

Adding it works ok in my nothing example that uses FTXUI with modules, and (module) examples build.
Also, maybe note that when consuming the modules with cmake, you do so with a `target_link_libraries(your_target PRIVATE ftxui-modules)` (or ftxui::modules - not sure why the alias)

diff --git a/cmake/ftxui_modules.cmake b/cmake/ftxui_modules.cmake
index 6c784a9..bce0e8b 100644
--- a/cmake/ftxui_modules.cmake
+++ b/cmake/ftxui_modules.cmake
 
@@ -6,6 +6,7 @@ add_library(ftxui-modules)
 
 target_sources(ftxui-modules
   PUBLIC FILE_SET CXX_MODULES FILES
+  src/ftxui/ftxui.cppm
   src/ftxui/component.cppm
   src/ftxui/component/animation.cppm
   src/ftxui/component/captured_mouse.cppm

tattwamasi:
Just a quick change to go ahead and enable the umbrella module discussed in issue #1083.

I want this PR to be minimal, but would think a nice follow-on PR would be to add an example or at least test for the `import ftxui;` case, or change the current module edit script to use it instead of the individual modules.

### DIFF
--- a/cmake/ftxui_modules.cmake
+++ b/cmake/ftxui_modules.cmake
@@ -6,6 +6,7 @@ add_library(ftxui-modules)
 
 target_sources(ftxui-modules
   PUBLIC FILE_SET CXX_MODULES FILES
+  src/ftxui/ftxui.cppm
   src/ftxui/component.cppm
   src/ftxui/component/animation.cppm
   src/ftxui/component/captured_mouse.cppm
--- a/doc/cpp20-modules.md
+++ b/doc/cpp20-modules.md
@@ -12,8 +12,24 @@ FTXUI experimentally supports
 compilation times and improve code organization. Each header has a
 corresponding module.
 
-**Example with CMake and Ninja**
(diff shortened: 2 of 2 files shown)


ArthurSonzogni (approved):

tattwamasi:
Thank you.  Do I need to do anything else for merging and closing?

PR #1094, opened 2025-08-10
Adds opt-in support for applications that need to read piped data from stdin while still receiving interactive keyboard input

HarryPehkonen:
## Summary

Adds opt-in support for applications that need to read piped data from stdin while still receiving interactive keyboard input.

## Changes

- Add `HandlePipedInput(bool enable)` method to `ScreenInteractive`
- Detects piped input and redirects stdin to `/dev/tty` when enabled
- POSIX-only feature with Windows/WASM stubs for compatibility
- Comprehensive test coverage
- Documentation in README

## Usage

```cpp
screen.HandlePipedInput(true);  // Enable before Loop()
screen.Loop(component);

Allows commands like cat data.txt | your_app to work with full keyboard interaction.

Backward Compatibility

- Feature is disabled by default - zero breaking changes
- Existing applications work exactly as before
- Cross-platform safe with appropriate guards

Testing

- Unit tests cover normal operation, edge cases, and failure modes
- Only runs on POSIX systems where feature is available

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,6 +378,22 @@ Several games using the FTXUI have been made during the Game Jam:
 - [smoothlife](https://github.com/cpp-best-practices/game_jam/blob/main/Jam1_April_2022/smoothlife.md)
 - [Consu](https://github.com/cpp-best-practices/game_jam/blob/main/Jam1_April_2022/consu.md)
 
+## Advanced Usage
+
+### Piped Input Support
+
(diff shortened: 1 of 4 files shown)


ArthurSonzogni:
Thanks! I like having this builtin.
Nice description.

I will merge it as soon as I can 😉

ArthurSonzogni:
**Question**

Would it make sense to enable this unconditionnally? I don't think anyone would ever need to opt-out of this.

Issue #1093, opened 2025-08-07
Vertical slider cannot be increased with Up key press


its-pablo:
See the following code snippet:

```
using namespace ftxui;

auto screen = ScreenInteractive::Fullscreen();
int value = 50;
auto slider_options =
    SliderOption<int>({.value = &value,
                       .min = 0,
                       .max = 100,
                       .increment = 1,
                       .direction = Direction::Up});
auto slider = Slider(slider_options);
screen.Loop(slider);
```

When running this simple slider example (adapted from https://github.com/ArthurSonzogni/FTXUI/blob/main/examples/component/slider.cpp), the slider value always decreases whether you press the Up or Down key. This is likely due to [this in slider.cpp](https://github.com/ArthurSonzogni/FTXUI/blob/994915dbb90271b5a96a3cf947784d6633b2d5d3/src/ftxui/component/slider.cpp#L78-L90):

```
  void OnUp() {
    switch (this->direction) {
      case Direction::Up:
        this->value() -= this->increment();
        break;
      case Direction::Down:
        this->value() += this->increment();
        break;
      case Direction::Left:
      case Direction::Right:
        break;
    }
  }
```

I believe Up press on `Direction::Up` should result in `+=` and Up press on `Direction::Down` should result in `-=`.

Note that the Down press works as expected, so the vertical slider cannot be increased with key presses at all.

ArthurSonzogni:
Thanks for reporting the issue!

PR #1099, opened 2025-08-19
Fix slider Up key press.

ArthurSonzogni:
The direction was inverted. It caused the inability to increase it.

Fixed:https://github.com/ArthurSonzogni/FTXUI/issues/1093


ArthurSonzogni:
@its-pablo FYI.

Thanks for reporting it!

PR #1106, opened 2025-08-27
Improved the installation method via Conan.

xiaoditx:
I've written a guide for the installation method via Conan. Although there might be some errors in what I've written—likely due to connectivity issues since I'm located in China (the error messages indicate connection failures)—I do have a verified version that works, even though it's a bit more cumbersome. I hope the format of what I've written is correct and can be helpful.


ArthurSonzogni:
Thanks!

PR #1107, opened 2025-08-31
Fixed bug in component/button example:

m00se-3:
- The '-1' and '+1' buttons now correctly increment and decrement.
 - Previously it was vice versa.


ArthurSonzogni (approved):

ArthurSonzogni:
Thanks!

PR #1111, opened 2025-09-05
Fix use of uninitialized cursor variable

bearzly:
The cursor_ variable was being default initialized, which causes undefined behaviour when accessing properties in
ScreenInteractive::Draw. This caused a crash when running with UBSAN.

```
ftxui/src/ftxui/component/screen_interactive.cpp:852:17: runtime error:
load of value 4195502944, which is not a valid value for type 'Shape'
```

This change causes cursor_ to be value initialized so that its members will be initialized to 0.

### DIFF
--- a/include/ftxui/screen/screen.hpp
+++ b/include/ftxui/screen/screen.hpp
@@ -60,7 +60,7 @@ class Screen : public Image {
       BarBlinking = 5,
       Bar = 6,
     };
-    Shape shape;
+    Shape shape = Hidden;
   };
 
   Cursor cursor() const { return cursor_; }


ArthurSonzogni:
Thanks! This is greatly appreciated. Maybe I should run the fuzzers with UBSAN too instead of only ASAN.

Could we instead initialize `Screen::Cursor::shape_` where is is declared?
```diff
class Screen {
  ...
  struct Cursor {
    int x = 0;
    int y = 0;

    enum Shape {
      Hidden = 0,
      BlockBlinking = 1,
      Block = 2,
      UnderlineBlinking = 3,
      Underline = 4,
      BarBlinking = 5,
      Bar = 6,
    };
-   Shape shape;
+   Shape shape = Hidden;
  };
  ...
}
```

bearzly:
Done, tested that this also fixes the issue.

ArthurSonzogni (approved):
Thanks! LGTM!

Issue #1108, opened 2025-09-04
Calling `vbox()` with  a collection else than `std::vector<Element>` silently fails


nbusser-sr:
## Description

In my code, I use a `std::dequeue<Element>`.

The code compiles when building a `vbox()` with it, but the SFINAE pattern on `Merge()` discards all the elements.

## Example

The following code returns a well constructed `vbox()`, containing a `text()`:
```cpp
Elements collection;
collection.push_back(text("TEST"));
vbox(collection);
```

This code however, returns an empty `vbox()`:
```cpp
std::dequeue<Element> collection;
collection.push_back(text("TEST"));
vbox(collection);
```

## Reason

When looking at the stack trace, the reason of the bug is clear:

https://github.com/ArthurSonzogni/FTXUI/blob/775ad9ce5e1208a7c575ca68ebd360cf6e4295a3/include/ftxui/dom/take_any_args.hpp#L42
https://github.com/ArthurSonzogni/FTXUI/blob/775ad9ce5e1208a7c575ca68ebd360cf6e4295a3/include/ftxui/dom/take_any_args.hpp#L35-L40
https://github.com/ArthurSonzogni/FTXUI/blob/775ad9ce5e1208a7c575ca68ebd360cf6e4295a3/include/ftxui/dom/take_any_args.hpp#L27-L33
https://github.com/ArthurSonzogni/FTXUI/blob/775ad9ce5e1208a7c575ca68ebd360cf6e4295a3/include/ftxui/dom/take_any_args.hpp#L12-L13

The template pattern expects to receive a `Elements`; i.e a `std::vector<Element>` in second argument, but it receives a `std::deque<Element>` instead, so we fall in the default case of the SFINAE: https://github.com/ArthurSonzogni/FTXUI/blob/775ad9ce5e1208a7c575ca68ebd360cf6e4295a3/include/ftxui/dom/take_any_args.hpp#L12-L25

## Expected behavior

I would expect the code either:
- To not compile if another collection than `Elements` is provided to `vbox`
- Properly allow `Merge()` with any type of collection

ArthurSonzogni:
Thanks for reporting!

I would be open to a PR fixing this. Both solution looks good to me. Would you like to submit a fix?

nbusser-sr:
I would be glad

ArthurSonzogni:
Merged! Thaks for your work!

PR #1117, opened 2025-09-17
Add `Merge()` specializations to support more `Element` containers

nbusser:
## Related issue

https://github.com/ArthurSonzogni/FTXUI/issues/1108

## Description

`Merge()` was previously only supporting `Elements` as a `Element` container.  
This PR adds specialization for:
- all the containers that matches the concept `std::ranges::range`
- `std::queue`
- `std::stack`

## Interrogation

I did not find a clear reference about the C++ standard that FTXUI claims to support.

@ArthurSonzogni Is it ok to use C++20 concepts here, or should FTXUI be compatible with older C++ standards? In that case I will change the PR content accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ Next
 ### Dom
 - Fix integer overflow in `ComputeShrinkHard`. Thanks @its-pablo in #1137 for
   reporting and fixing the issue.
+- Add specialization for `vbox/hbox/dbox` to allow a container of Element as
+  as input. Thanks @nbusser in #1117.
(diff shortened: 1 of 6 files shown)


ArthurSonzogni:
> @ArthurSonzogni Is it ok to use C++20 concepts here, or should FTXUI be compatible with older C++ standards? In that case I will change the PR content accordingly.

Yes, I am not going to drop C++17 users soon. Could you please wrap the C++20 parts behind:
```
#if __cplusplus >= 202002L
...
#endif
```

PR #1122, opened 2025-09-21
Fixed SliderOption compilation error when compiling with MSVC

DanielPetk:
## Problem
I've ran into a problem when trying to compile from source using MSVC from CMAKE's 'Visual Studio 17 2022' generator. I'm not sure if this happens in other circumstances, but this is where I encountered this issue. When trying to compile, I get the following error:
```
include\ftxui\component\component_options.hpp(227,28): error C2760: syntax error: ')' was unexpected here; expected 'expression'
```
The struct is as follows:
```cpp
template <typename T>
struct SliderOption {
  Ref<T> value;
  ConstRef<T> min = T(0);
  ConstRef<T> max = T(100);
  ConstRef<T> increment = ((max)() - (min)()) / 20;
  Direction direction = Direction::Right;
  Color color_active = Color::White;
  Color color_inactive = Color::GrayDark;
  std::function<void()> on_change;  ///> Called when `value` is updated.
};
```
The culprit is this line:
```cpp
ConstRef<T> increment = (max() - min()) / 20;
```
## Solution
My assumption is that MSVC sees `max()` and `min()` as function calls calling the standard library's max and min function instead of calling `operator()` on the `min` and `max` members. I solved this by simply adding parentheses:
```cpp
ConstRef<T> increment = ((max)() - (min)()) / 20;
```
This is my first pr on GitHub, so forgive me for any mistakes.

### DIFF
--- a/include/ftxui/component/component_options.hpp
+++ b/include/ftxui/component/component_options.hpp
@@ -226,7 +226,7 @@ struct SliderOption {
   Ref<T> value;
   ConstRef<T> min = T(0);
   ConstRef<T> max = T(100);
-  ConstRef<T> increment = (max() - min()) / 20;
+  ConstRef<T> increment = ((max)() - (min)()) / 20;
   Direction direction = Direction::Right;
   Color color_active = Color::White;
   Color color_inactive = Color::GrayDark;


ArthurSonzogni:
Hello @DanielPetk 

Thank you for your pull request. I appreciate you finding and reporting this issue. I’d like to understand a bit more about what caused this compilation error for you.

You've correctly noted that the compiler gets confused by min() and max(). There are a few different reasons why this might happen:

- A macro, such as those defined in windows.h. We recently added <ftxui/util/warn_windows_macro.hpp> to help diagnose this specific issue.
- A using namespace std; directive.
- A name collision with std::min and std::max in the global scope.

Could you tell me what the specific cause was in your case? This will help me confirm the best way to handle this issue.

Thank you again for your contribution.

danpetk:
It is most definitely the windows.h header thats the problem. I'm not sure why I considered a collision with std::min and std::max when clearly there is no using namespace std, I guess that's what i get for doing stuff at 4am 🙃.

 I think I read in a different PR that you choose not to change anything in this case. Sorry about that!

ArthurSonzogni:
Thanks for the confirmation!

Yes, we don't believe it's sane to avoid using the min and max keywords just because another library has defined a macro for them. The diagnostic we added previously should allow developers to cope with the `windows.h` issue.

PR #1137, opened 2025-10-19
Mitigate integer overflow in ComputeShrinkHard

its-pablo:
The multiplication in ComputeShrinkHard has the potential to overflow when very large elements are present inside of a vbox. To mitigate this while keeping the same behavior (multiply first), we can perform a saturating multiply before the division.

~~Alternatively, we could just do the division first.~~ 🫣

### DIFF
--- a/src/ftxui/dom/box_helper.cpp
+++ b/src/ftxui/dom/box_helper.cpp
@@ -10,6 +10,7 @@
 namespace ftxui::box_helper {
 
 namespace {
+<<<<<<< HEAD
 // Multiplication that clamps on overflow. Can be replaced by std::mul_sat
 // in C++26.
 int mul_sat(int a, int b) {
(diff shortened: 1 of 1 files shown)


ArthurSonzogni:
Thanks!

> The multiplication in ComputeShrinkHard has the potential to overflow when very large elements are present inside of a vbox. To mitigate this while keeping the same behavior (multiply first), we can perform a saturating multiply before the division.7

Did you experienced a bug for real, or is it an hypothetical bug with a degenerated reproducer?

> Alternatively, we could just do the division first.

`*(int,int)` and `/(int, int)` do not commute, and this would gives the wrong result.

its-pablo:
Bug for real! I am working on a virtual view port for very large text files within FTXUI (100s of thousands of lines!) and this int overflow was keeping my view pegged to the first element of the vbox for elements beyond a certain size (AKA if too many lines in log file). For files under some line count the viewer was working fine. Hooked up the debugger to figure out what was up and this is were the sizes were going wonky!

> Alternatively, we could just do the division first.

# 🫣 

Well that's an embarrassing statement from me!

its-pablo:
I can put together a repro example, just not at liberty to share the particular example I am working with locally!

its-pablo:
Ugggggh... I apologize! I didn't PR with the tip of my local changes.

its-pablo:
Apologies @ArthurSonzogni , I somehow managed to mess up my git tree when PRing. I closed this out and opened a fresh draft. See https://github.com/ArthurSonzogni/FTXUI/pull/1138

PR #1138, opened 2025-10-19
Mitigate integer overflow in ComputeShrinkHard

its-pablo:
Please see: https://github.com/ArthurSonzogni/FTXUI/pull/1137

> The multiplication in ComputeShrinkHard has the potential to overflow when very large elements are present inside of a vbox. To mitigate this while keeping the same behavior (multiply first), we can perform a saturating multiply before the division.

I noticed I accidentally messed up my git tree and PR'd with the wrong change set. To actually fix this behavior we have to perform the full size calculation into the long long otherwise we clamp too early and still encounter the bad behavior.


ArthurSonzogni:
Thanks!

its-pablo:
Thanks for the quick merge! Can now point my FetchContent back at you! 😀

PR #1015, opened 2025-03-25
Add modules support

mikomikotaishi:
I have tried making a C++ modules version for FTXUI, which is built if CMake version is above 3.28.

PR #1146, opened 2025-11-10
Use module partitions instead of full modules

mikomikotaishi:
Follow-up to #1015. This pull request replaces the full modules that represent headers, with partitions, to emphasise the belonging of the header to the module. This should hopefully provide a speedup to compilation, and confuse users less by aggregating the usable modules into a smaller set.

### DIFF
--- a/doc/cpp20-modules.md
+++ b/doc/cpp20-modules.md
@@ -73,38 +73,6 @@ are available:
 
 - `ftxui`
     - `ftxui.component`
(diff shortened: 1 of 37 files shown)


ArthurSonzogni (approved):
Thanks!

LGTM % remove the partitions from the documentation.

ArthurSonzogni:
Thanks!

Issue #1153, opened 2025-11-26
String view support


mikomikotaishi:
Could we make it so that functions like `ftxui::text()` and others that accept `std::string` also accept `std::string_view` as well?

razor85:
I came here because I was about to open a new issue with this exact same request. +1 please.
EDIT: Just noticed you linked a PR that adds string_view support but that doesn't help because it will always keep a copy of the string, making the changes useless memory-wise. Ideally the lib should either keep a string cache outside of its structures, make the user responsible for their strings memory or store something like a std::variant<std::string_view, std::string> (in a class) with the correct chosen option (from the class constructor).

Epixu:
On the topic of `string_view`, I have a suspicion that paragraphs can also benefit a lot from only interfacing the contents of a main `string` (or `string_view`), rather than splitting the text into individual `Node`s on the heap

mikomikotaishi:
> EDIT: Just noticed you linked a PR that adds string_view support but that doesn't help because it will always keep a copy of the string, making the changes useless memory-wise.

I only made a quick overload for it that just copies `string_view` into `string`, if @ArthurSonzogni knows of other places to update the API with `string_view` support I'll eventually try and make a more proper PR for this.

mikomikotaishi:
@ArthurSonzogni Also, do you know if things like `Component` support newlines in strings? When I connect one with `Menu()` and `Vector<String>`, it prints the `\n` literally rather than inserting a newline.

If I'm doing it wrong, could you let me know what the correct thing is to do? Otherwise, could we try to push for this before release v7?

ArthurSonzogni:
Adding new line support for text() would magically add support for newline in every component.
I think I am open to it. Do you want to try submitting a PR for this?

We won't support line wrapping, as we do the layout in a single pass, but that can still be useful.

mikomikotaishi:
I can try to give it a shot but I'm not too familiar with the implementation of text(), but I will try to give it a look soon

mikomikotaishi:
OK, take a look at #1215.

PR #1154, opened 2025-11-26
Add string view overloads

mikomikotaishi:
The function `ftxui::text()` lacks overloads for `std::string_view`. This pull request adds support for it. I opened an issue #1153 for a general request, but I'm not sure if there are others that need `string_view` overloads.

I proposed this so that I can avoid having to manually cast a `string_view` to a `string` in my own code, and I am sure others may find this useful as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,11 @@ Next
 - Remove dependency on 'pthread'.
 - Bugfix: Bazel target @ftxui is now visible. Thanks @dskkato in #1157.
 
+### General
+- Breaking. Move to `std::string_view` instead of `const std::string&` where
+  applicable. This yields better interoperability with string literals and
+  avoids unnecessary copies. Thanks @mikomikotaishi for PR #1154
+
 ### Component
 - Feature: POSIX Piped Input Handling.
   - Allows FTXUI applications to read data from stdin (when piped) while still receiving keyboard input from the terminal.


mikomikotaishi:
@ArthurSonzogni Do you know any other places in the library where `string` is used? I would like to increase the amount of `string_view` overloads if possible, I am making something that uses `string_view` a lot and would like to be able to pass it more elegantly without having to construct `string` from it

ArthurSonzogni:
Thanks!

I don't know why I wasn't using `std::string_view` instead of `const std::string&` or `std::string`. Maybe I supposed it was C++20 and I want to keep C++17 support for some time.

I am going to update your patch to convert everything.

mikomikotaishi:
Thanks, such an update will make my life a lot easier. (The current patch I made is a lazy patch that just casts it to `string`, but hopefully we can remake this patch to actually leverage `string_view`'s cheap cost rather than just construct `string` from it anyway.)

ArthurSonzogni:
Thanks!

I added other string_view

mikomikotaishi:
I think we should avoid constructing `string` from `string_view`, and try and leverage `string_view`'s non-ownership of the data, as this is cheaper and a better use of `string_view` itself.

mikomikotaishi:
Actually, forget my previous comment. I don't know of another way to do this as the class itself is storing `string`, which must be constructed from the `string_view`.

mikomikotaishi:
@ArthurSonzogni Do you have any thoughts about what to do with `string_view` or how to leverage it the most effectively here?

ArthurSonzogni:
Thanks!

I don't think I know how to leverage std::string_view and avoid storing the strings.

Sure, we can store a std::string_view, but this would tie the lifetime of the object with the callers. This is possible if the users of FTXUI are understanding C++ and are always good enough to avoid memory safety issues.

One of the reasons I am using `std::shared_ptr` instead of `std::unique_ptr` for all of the Element and Component is that developers were having too many memory safety issues when dealing with memory more directly.

ArthurSonzogni:
Note: I ran benchmarks to estimate the performance improvement. We got a regression of ~2.97%.

I guess I was either unlucky (need more run), or this is real. I believe the better ergonomics are worth spending a 3% regression.

mikomikotaishi:
Maybe it's worth keeping the `const string&` signatures which use `std::move`, along with `string_view`?

mikomikotaishi:
@ArthurSonzogni I think we still need to update the ones in `component.hpp`, many are still taking `ConstStringRef` rather than `string_view`. We should also update `ConstStringListRef` to accept `string_view`, or just create a new overload for `const vector<string_view>>&`.

ArthurSonzogni:
I added: 942ab6a82d8dd555e7b8961a2e2dcab8c6bac54e
Does that make sense to you?

mikomikotaishi:
Great news!

PR #1156, opened 2025-11-29
Added the translation link

xiaoditx:
I simply added the link and also made some minor changes to the previous todo list I had written.

The translation provided by LLM is not very accurate. In some parts (at least in the simplified Chinese version), they were not translated. I wanted to correct it but I wasn't sure how to do it while ensuring that the subsequent translations wouldn't overwrite. So, I decided not to fix the issues for now.

By the way, it seems that [https://github.com/ArthurSonzogni/ftxui-translations/tree/fr] has not been translated. What's the reason for this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,12 @@
   <a href="https://arthursonzogni.github.io/FTXUI/examples/">Examples</a> .
   <a href="https://github.com/ArthurSonzogni/FTXUI/issues">Request Feature</a> ·
   <a href="https://github.com/ArthurSonzogni/FTXUI/pulls">Send a Pull Request</a>
+  
+  <br/>
+  <a href="https://github.com/ArthurSonzogni/">English</a> |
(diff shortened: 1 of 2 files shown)


ArthurSonzogni:
> By the way, it seems that [https://github.com/ArthurSonzogni/ftxui-translations/tree/fr] has not been translated. What's the reason for this?

I think it does. This is in French. I am french.


> I wanted to correct it but I wasn't sure how to do it while ensuring that the subsequent translations wouldn't overwrite. So, I decided not to fix the issues for now.

I made the script so that the LLM will compare the diff in between new versions of FTXUI and the previously translated version. Then only translate the new diff. This is subject to LLM interpretation, but I am confident this should override manual translations.

If you are interested, it means you could fetch the related branch and update the parts needed from your own translations.

xiaoditx:
> I think it does. This is in French. I am french.

Oh yes, I just noticed. It seems I'm really not very sensitive to French, though it's probably more because I wasn't browsing carefully enough.

As for the manual corrections, I'm already working on them now, so it shouldn’t take long.

PR #1157, opened 2025-12-04
fix(bazel): ensure FTXUI is publicly accessible and add external smoke test

dskkato:
## Summary
This PR fixes the Bazel build configuration to make FTXUI properly consumable from external workspaces using Bzlmod, and adds integration tests to prevent regressions.

## Problem
The previous alias target for `:ftxui` did not properly re-export headers from the underlying modules (`component`, `dom`, `screen`), causing compilation failures when external projects tried to consume as `@ftxui//:ftxui`:

```sh
error: module //:smoke does not depend on a module exporting 'ftxui/dom/elements.hpp'
```

## Solution

1. Replace alias with ftxui_cc_library

Changed the top-level `:ftxui` target from a simple `alias` to a `ftxui_cc_library` that explicitly re-exports all public headers using `glob(["include/ftxui/**/*.hpp"])`. This ensures downstream consumers can access all FTXUI headers through the single `@ftxui//:ftxui` dependency.

2. Add External Integration Test
Created `bazel_integration` as a minimal external workspace that:

* Uses Bzlmod with `bazel_dep(name = "ftxui")`
* Overrides the dependency to test the local version
Includes a smoke test that exercises all three modules
* Validates that `@ftxui//:ftxui` works as a single dependency

3. Add CI Validation
Added a new CI job that builds the smoke test using:

```sh
bazel build //:smoke --enable_bzlmod --override_module=ftxui=..
```

This prevents future regressions in external visibility or API accessibility.

## Testing

```sh
cd bazel_integration
bazel build //:smoke --enable_bzlmod --override_module=ftxui=..
./bazel-bin/smoke  # Successfully runs interactive UI
```

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -7,4 +7,4 @@ build --enable_bzlmod
 build --enable_platform_specific_config
 build:linux --cxxopt=-std=c++20
 build:macos --cxxopt=-std=c++20
-build:windows --cxxopt=-std:c++20
+build:windows --cxxopt=/std:c++20


dskkato:
It seems that I need to spend more time to resolve the build error other than `Bazel, g++, ubuntu-latest`. I'll make this PR open once they are resolved.

image


https://github.com/dskkato/FTXUI/actions/runs/19921383118/job/57110656658?pr=1

Issue #495, opened 2022-10-05
can't input non-ascii character on Input component


twinkles-twinstar:
I don't know if this is expected behavior or bug, FTXUI supports UTF-8 output, but it seems to only support ASCII input?.
I tried to enter non-ascii characters (e.g. CJK characters, full-width commas, emoji) in the input component, but none of them entered correctly into the input component, and I also tried copying a non-ascii text from the outside and then copying it in the input component, but still could not enter it correctly.

https://user-images.githubusercontent.com/37923060/194156167-12c998f4-316b-46e1-9418-27336d29bcca.mp4

TwinKleS-C:
I think it may be that non-ascii bytes have an effect on the operation of ftxui, and in some of my other tests, entering non-ascii text may cause control strings to appear in the interface or even crash the program

TwinKleS-C:
BTW, I don't know if this is the reason: in the terminal under Windows, even if the codepage is set to utf-8 (using SetConsoleCP and SetOutputConsoleCP), you can't get the normal utf-8 text through std::cin, you must use ReadConsole api
The following is an example of how I can get a line of UTF-8 text from a Windows terminal :
```
auto input (
) -> std::string {
	#if defined M_system_windows
	auto text_16 = std::array<char16_t, 0x1000>{};
	auto length = DWORD{};
	auto state = ReadConsoleW(GetStdHandle(STD_INPUT_HANDLE), text_16.data(), static_cast<DWORD>(text_16.size()), &length, nullptr);
	assert_condition(state);
	auto text_8 = utf16_to_utf8(std::u16string_view{text_16.data(), length - 2});
	return std::string{std::move(reinterpret_cast<std::string &>(text_8))};
	#endif
	#if defined M_system_linux || defined M_system_macos || defined M_system_android || defined M_system_ios
	auto text = std::string{};
	std::getline(std::cin, text);
	return text;
	#endif
}
```

ArthurSonzogni:
Hello!

Sorry, support for Microsoft's terminal is not as good as the one in regular terminal.
I don't really have an easy access to a computer with Microsoft's Windows OS, so it is kind of hard for me.
Experimental support was kindly added by @mauve some time ago.

My understanding of the issue is that we do an incorrect cast here:
https://github.com/ArthurSonzogni/FTXUI/blob/master/src/ftxui/component/screen_interactive.cpp#L106

[...]

PR #1160, opened 2025-12-06
Fix UTF-16 surrogate pair handling on Windows input

739C1AE2:
This PR addresses the remaining issue in #495 where non-BMP characters (e.g., emojis) could not be entered.
﻿
The previous fix (#538) for UTF-16 input on Windows did not handle UTF-16 surrogate pairs correctly, causing non-BMP characters to be broken or unenterable in Input components.


ArthurSonzogni (approved):
Excellent! Thanks you!

This is particularly useful, because I can't easily test Windows compared to Linux.

ArthurSonzogni (approved):

PR #1164, opened 2025-12-14
Update nix lock and make nix package follow the current version

Rucadi:
With nix, usually you want to define the version of the code to be the one of the git-branch you are taking it, in this case, it was hardcoded to be version v4.0.0 (which is already quite old). 

For example, the usual is to do something like:

nix shell github:ArthurSonzogni/FTXUI

to get the latest version.

You can also get an specific commit via:

nix shell github:ArthurSonzogni/FTXUI/183a426efa93e7a1b25730e85e194246523ef1c8

(Which also works with branches or tags)

this fixes it for new releases, at the same time that updating the lockfile makes it use the newer versions of the libraries and tooling.

Check that nix builds are still working:  
image

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1694529238,
-        "narHash": "sha256-zsNZZGTGnMOf9YpHKJqMSsa0dXbfmxeoJ7xHlrt+xmY=",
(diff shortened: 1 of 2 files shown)


ArthurSonzogni (approved):

ArthurSonzogni:
Excellent! Thanks @Rucadi !

PR #1171, opened 2025-12-27
Fixed two potential issues

xiaoditx:
I made changes to two potential issues:

1. doc/header.html – there seemed to be a JavaScript syntax error inside.  
2. CMakeList – it appeared to include the same file repeatedly (might affect efficiency?To be honest , I'm not really sure,).


ArthurSonzogni:
Thanks! @xiaoditx

PR #1176, opened 2026-01-26
Closes #1150

### LINKED ISSUE #1150: Dropdown component does not check intersections
When using a dropdown and a button inside a container the dropdown will handle the onclick event even when the button is clicked.
As far as i could check, this is because of a missing intersection check inside the dropdown component.
The button checks if the event is meant for the component but the dropdown component does not.

techsupport-noah:
Closes #1150


ArthurSonzogni (approved):
Thanks for fixing the bug!

Issue #1179, opened 2026-02-01
Add focused_entry() = selected(); in MenuBase constructor


CoinFuMasterShifu:
https://github.com/ArthurSonzogni/FTXUI/blob/3317c1dd482dccc18af7bdc2c2eadbc123e9a6a7/src/ftxui/component/menu.cpp#L71
should be replaced by:

```c++
  explicit MenuBase(const MenuOption& option) : MenuOption(option) {
        focused_entry() = selected();
  }
```
because otherwise the first element will be focused even if some other shall be preselected on construction based on `option`.

ArthurSonzogni:
Thanks!
Yes, this sounds like a positive change. Would you like to submit a PR for this?

CoinFuMasterShifu:
I pulled your library into my project and have a different build system (meson), I don't use cmake. I can make this PR based on the above shown replacement and it compiles in my case but I did not compile this repo as cmake project to verify correctness. I will make the PR when I come back.

PR #1182, opened 2026-02-03
Set focused entry in MenuBase constructor

CoinFuMasterShifu:
Initialize focused entry to selected entry in MenuBase. Fixes https://github.com/ArthurSonzogni/FTXUI/issues/1179


ArthurSonzogni:
Thanks you!

PR #1189, opened 2026-02-15
Refactor Color::Interpolate

Rucadi:
I want to contribute a A small refactor which should make the code a little bit cleaner using c++17  structured bindings: https://en.cppreference.com/w/cpp/language/structured_binding.html

The code should be equivalent.


ArthurSonzogni:
Thanks for your PR. I like it!

PR #1195, opened 2026-02-21
fix(flexbox): correct child index skipping when line has no selection

739C1AE2:
Fix a bug in Flexbox::Select() where child indices get misaligned when skipping lines without selection.

When iterating through lines, if a line's box has no intersection with the selection, the code `continue`s without incrementing the child index `i`. This causes subsequent lines to access wrong `children_` elements.


ArthurSonzogni:
Thanks @739C1AE2 !

PR #1198, opened 2026-02-26
Fix some cursor issues

### LINKED ISSUE #1196: Input Component Cursor Positioning and Scroll Reset Issues
## Overview
I have identified several regressions in the `Input` component that seem to have been introduced during the focus logic refactoring in PR #978 :

1. **Incorrect cursor position when input is empty**
   - When the input content is empty and focused, the cursor appears at the end of the input box (or at the bottom-right corner of the screen when `FTXUI_MICROSOFT_TERMINAL_FALLBACK` is defined), instead of the expected starting position.
   - This causes IME preview windows to appear in the wrong location, affecting CJK character input.

Image

Image

2. **Scroll reset when input loses focus**
   - When a non-empty input loses focus, the previous scroll position is lost and the view scrolls back to the beginning.
   - This was previously mentioned in [this comment](https://github.com/ArthurSonzogni/FTXUI/pull/978#issuecomment-2699131318).

https://github.com/user-attachments/assets/8dce8db1-5462-4850-9426-fa096ebf571e

3. **Cursor position off by one for wide characters**
   - When the input contains wide characters (e.g., CJK characters), the cursor appears on the second half of the character instead of the first half (though some terminals may auto-correct this).

Image

Image

## Analysis

- Commit **dea4aa3** changed cursor coordinates from `x_min`/`y_min` to `x_max`/`y_max`. This causes the cursor to always be placed at the bottom-right corner of the focused area (responsible for issues 1 and 3).
  https://github.com/ArthurSonzogni/FTXUI/blob/a9c0d17eeceecde9b20e3654cd0722811390a4ae/src/ftxui/dom/node.cpp#L147-L151

- Commit **be73922** changed the decorator for unfocused state from `focus` to `nothing` in the Input component. This results in no focus area at all when the component is not focused (responsible for issue 2).
  https://github.com/ArthurSonzogni/FTXUI/blob/a9c0d17eeceecde9b20e3654cd0722811390a4ae/src/ftxui/component/input.cpp#L101-L103

## Proposed fixes

1. **Revert cursor coordinates**: Change the cursor coordinates in `Render()` back to `x_min`/`y_min`
2. **Restore focus area for unfocused state**: Change the `nothing` decorator back to `focus` for the unfocused state in the Input component
3. **Use the same focused variable for empty content**: Apply the same `focused` variable to the empty content branch

I'm not entirely certain about the original intentions behind these changes, and whether implementing these fixes might introduce new issues or break existing behavior that depends on the current implementation.

739C1AE2:
Fixes #1196 and improves non-ASCII support in password mode.

Key Fixes:

 - Cursor & Focus: This fixes the IME alignment and the scroll-reset issue reported in #1196.

 - Non-ASCII Passwords: Updated password masking to use `GlyphCount` instead of byte size. 

 - Flicker Reduction: Added `\033[?25l` before redrawing in `ScreenInteractive` to mitigate cursor ghosting during rapid updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ Next
 ### Component
 - Fix `Input` cursor positioning and scroll stability. See #1196. Thanks @739C1AE2.
 - Fix `Input` support for non-ASCII characters in password mode. See #1196. Thanks @739C1AE2.
+- Performance: Mitigate cursor flickering during redraw in `ScreenInteractive`. See #1196. Thanks @739C1AE2.
 - Feature: POSIX Piped Input Handling.
   - Allows FTXUI applications to read data from stdin (when piped) while still receiving keyboard input from the terminal.
   - Enabled by default.
--- a/src/ftxui/component/screen_interactive.cpp
+++ b/src/ftxui/component/screen_interactive.cpp
@@ -916,6 +916,9 @@ void ScreenInteractive::Draw(Component component) {
       break;
   }
 
+  // Hide cursor to prevent flickering during reset.
+  std::cout << "\033[?25l";
+
   const bool resized = frame_count_ == 0 || (dimx != dimx_) || (dimy != dimy_);
   ResetCursorPosition();
   std::cout << ResetPosition(/*clear=*/resized);


ArthurSonzogni:
Excellent! Thanks @739C1AE2 !

The patch looks great to me at first sight. I am going to take some time later to review it carefully, test it locally and update the CHANGELOG.md file.

PR #1202, opened 2026-02-28
Big rename

ArthurSonzogni:
This commit renames several core components of the FTXUI library to better reflect their purpose and improve clarity. The following changes are included in this commit:
- `Pixel` is renamed to `Cell` to better represent its role as the basic unit of the terminal screen.
- `Image` is renamed to `Surface` to more accurately describe its function as a drawable area that can be rendered to the screen.
- `ScreenInteractive` is renamed to `App` to better convey its role as the main interface for creating interactive applications with FTXUI.


ArthurSonzogni:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/ArthurSonzogni/FTXUI/pull/1202#pullrequestreview-3871326933)

ArthurSonzogni:
Hi everyone,

CC: @markmandel @mikomikotaishi @dskkato @739C1AE2 @HarryPehkonen @zozowell @its-pablo @Smail @Sckab @nbusser @kenReneris @killerdevildog

I am thinking about releasing a new version of FTXUI soon.

Before doing so, there is something I have wanted to do for years: rename some core classes whose previous names I have grown to regret. Specifically, in PR #1202, I am making the following changes:

* `Pixel` -> `Cell`
* `Image` -> `Surface`
* `ScreenInteractive` -> `App`

I am CCing you all to get the room temperature on this change. I would also love to hear if you have any suggestions or calls for other class renames!

Please note that this PR is fully backward compatible with the old names. The old names will remain supported for a couple of versions and will be kept around as long as maintaining them isn't a problem.

Let me know what you think!

markmandel:
No issue on my end - my only suggestion would be to bump the major semver number to indicate a breaking change.

Arguably, then you also don't need to maintain backward compatibility.

killerdevildog:
One thought on compatibility: in my experience, a smoother path is to
*deprecate* the old names first (keep aliases/shims + warnings for one
release line), then *remove* them in a following release. For example: keep
deprecated aliases through the next minor/patch releases, and fully remove
them at the next major bump (or after a clearly stated deprecation window).

That keeps downstreams from getting hard-broken immediately while still
moving the project forward.

— Quaylyn

mikomikotaishi:
I agree, for the time being use a `[[deprecated()]]` marker.

markmandel:
Apparently my viewpoint is controversial - and that's okay 😄 

In my experience, people wait until the end of the deprecation window to switch over, so it becomes no real difference from switching immediately or not -- but definitely not against a deprecation window if the community desires! 😄

mikomikotaishi:
I think since this is the only release with the new names it makes sense to first use a deprecation marker before fully committing to phasing out the old names, as long as there's some warning to users that there is a renaming it would be appreciated

mikomikotaishi:
I'll give some more suggestions for potential renames in a bit. But for the time being:
- Please consider using `[[deprecated("Deprecation message")]]` on the old names so that there's an opportunity for users to be aware these names are deprecated
- Could we potentially integrate some of the library components? Among these are [ftxui-gridcontainer](https://github.com/mingsheng13/grid-container-ftxui), [ftxui-ip-input](https://github.com/mingsheng13/ip-input-ftxui), [ftxui-image-view](https://github.com/ljrrjl/ftxui-image-view), [ftxui-navigation-tree](https://github.com/Appisolato/navigation-tree-ftxui), and [MarkdownFTXUI](https://github.com/zvasilev/MarkdownFTXUI). I don't know whether the authors are willing to integrate it into the main project, but a lot of them seem like very useful components that would be worth integrating. Perhaps just food for thought for the next release.
- A lot of functions, like `ftxui::CatchEvent`, or `ftxui::Menu`, return `ftxui::Component` internally, though it would be kind of nice if it were easier to distinguish by a type name when something should be `Menu` and when something else, like `Renderer`. Maybe there is a reason for this, but if there is it eludes me at the moment.

nbusser:
Thank you for cc.ing me. I think the changes totally make sense. No problem on my end 🙂

ArthurSonzogni:
Thanks for your input!

We can keep the compatibility names for a couple of years, or even indefinitely, as long as we don't hit problem maintaining them. With the new document and examples, this will allow  developers to switch smoothly.

About `[[deprecated("Deprecation message")]]`, yes I probably should at some point, but this would currently fail to build due to `-Werror`. I need to update some flags to allow this.

PR #1203, opened 2026-02-28
fix: gaugeUp/gaugeDown fill full allocated width when decorated with flex

Ardet696:
# Problem

  gaugeUp and gaugeDown always render in a single character column
  box_.x_min, ignoring any extra width allocated to them. This means
  that even when a width-growing flex decorator is applied, the gauge
  renders as 1 column wide regardless of available space:

  ## Fix

  RenderVertical now loops over all columns from box_.x_min to
  box_.x_max, drawing the same fill level across the full allocated
  width. 
  The invert pass is kept outside the column loop so the XOR
  toggle is applied exactly once per pixel.

  This mirrors how RenderHorizontal already fills its full allocated
  height when given extra vertical space.

 ### Motivation

  This came up building a spectrum analyser bar display,  multiple
  gaugeUp elements in an hbox with xflex_grow to share the
  available width. The fix makes vertical gauges behave symmetrically
  with horizontal ones.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,8 @@ Next
 - Bugfix: In the gridbox, add the children to the tree so that the default
   behaviors inherited from Node are correctly implemented. Thanks KenReneris for
   #1070.
+- Update: The `gauge` in a flexible now takes the available space in the
+  opposite direction. Thanks @Ardet696 in #1203.
 
 ### Screen
 - Performance: Optimize `Screen::ToString()`, `Color::Print()` and


ArthurSonzogni:
Thanks!

Should we do the same for horizontal gauge?

Ardet696:
> Thanks!
> 
> Should we do the same for horizontal gauge?

Yeah sure! I will update the PR.

Ardet696:
Hello, can we proceed here?
Is there testing needed?

ArthurSonzogni:
Thanks!

PR #1221, opened 2026-04-08
Consolidate code into a single module to reduce compile times

mikomikotaishi:
We can greatly lower the compilation time by decreasing the number of partition translation units, by moving all of the symbol exports into `ftxui:*` modules instead of `ftxui.**:*`. Furthermore, because all modules are built anyway, it is a better idea to just provide only the full `ftxui` module rather than leave it up to the user to import three separate `ftxui.*` modules.


ArthurSonzogni:
Thanks!

   - Consolidated Modules: Switched from partitions back to named modules (ftxui.screen, ftxui.dom, ftxui.component, and ftxui.util).
   - GCC Fix: Corrected the import order in generated examples to avoid GCC internal compiler errors.
   - Removed Deprecations: Removed the accidentally added [[deprecated]] attributes.
   - Documentation: Restored the list of submodules in doc/cpp20-modules.md.
   - Changelog: Added an entry for PR #1221.


I still want users to be able to import only ftxui.screen or ftxui.dom, because not all user want to build interactive components.

I reverted the [[deprecated...]], because this was unrelated and I don't want to use them, as it would cause compile errors on embedders.

I fixed some issues preventing GCC to build/use modules.

mikomikotaishi:
> I still want users to be able to import only ftxui.screen or ftxui.dom, because not all user want to build interactive components.

But enabling building modules will build all modules anyway.

mikomikotaishi:
@ArthurSonzogni Since FTXUI now offers the full library as a single header `<ftxui/ftxui.hpp>`, why not just offer only one module `ftxui` and make `ftxui.component`, `ftxui.dom`, etc. into partitions (`ftxui:component`, `ftxui:dom`, etc.)? Any thoughts?

Even if the build is configured to not build the screen, dom, etc. parts of the library, these could be put behind an build macro so that those modules are built and exported only if the build setting exists.

Issue #1261, opened 2026-05-09
FTXUI interactive starts too slow


kghost:
This single line cost 0.5s, which is boaderline acceptable for a small interactive program.

``` c++
  _FtxuiLoop = std::make_unique<::ftxui::Loop>(&_Screen, component);
```

Here is my main function

``` c++
  using namespace ::ftxui;

  std::cerr << std::chrono::system_clock::now() << " profile 1" << std::endl;

  auto renderer = Renderer([&] {
    if (_ActiveView) {
      return _ActiveView->Render();
    }
    return text("Loading...");
  });

  std::cerr << std::chrono::system_clock::now() << " profile 2" << std::endl;

  auto component = CatchEvent(renderer, [&](Event event) -> bool {
    if (event == Event::Character('q')) {
      Stop();
      return true;
    }
    if (_ActiveView && _ActiveView->OnEvent(event)) {
      _Screen.PostEvent(Event::Custom);
      return true;
    }
    return false;
  });

  std::cerr << std::chrono::system_clock::now() << " profile 3" << std::endl;
  // Default view
  _ActiveView = std::make_shared<MyProgram>(_Tick);
  std::cerr << std::chrono::system_clock::now() << " profile 4" << std::endl;
  _FtxuiLoop = std::make_unique<::ftxui::Loop>(&_Screen, component);

  std::cerr << std::chrono::system_clock::now() << " profile 5" << std::endl;

  // Run twice to ensure the screen is properly initialized and rendered before entering the main loop
  _FtxuiLoop->RunOnce();

  std::cerr << std::chrono::system_clock::now() << " profile 6" << std::endl;
  _ActiveView->OnTerminalSizeChange(); // Trigger initial size update
  std::cerr << std::chrono::system_clock::now() << " profile 7" << std::endl;
  _Screen.PostEvent(::ftxui::Event::Custom);
  std::cerr << std::chrono::system_clock::now() << " profile 8" << std::endl;
  _FtxuiLoop->RunOnce();

  _Loop.Run();

  // Cleanup
  _FtxuiLoop.reset();
```

This is the timestamp result:

```
2026-05-09 16:47:49.987719508 profile 1
2026-05-09 16:47:49.987818505 profile 2
2026-05-09 16:47:49.987840652 profile 3
2026-05-09 16:47:49.987872189 profile 4
2026-05-09 16:47:50.495096272 profile 5
2026-05-09 16:47:50.523920575 profile 6
[...]

PR #1262, opened 2026-05-09
Fix starts slow on Windows terminal (#1261)

kghost:
This PR applies the local fix for Windows terminal startup latency by reordering terminal capability queries and waiting for the terminal capabilities response before continuing.


ArthurSonzogni (approved):

ArthurSonzogni:
Excellent! Thanks for the bisection, the analysis, and the fix!

PR #1276, opened 2026-05-22
component: Add CatchSignals option to App

kghost:
### Description
This pull request introduces the `CatchSignals` option to the `App` class. 

By default, FTXUI installs signal handlers for standard terminate and crash signals (`SIGTERM`, `SIGSEGV`, `SIGINT`, `SIGILL`, `SIGABRT`, and `SIGFPE`) to safely clean up and restore the terminal state before exit.

However, in many host applications (such as those with dedicated crash-reporting, stack-tracing, or customized signal handling), having FTXUI intercept these signals can conflict with standard error handling. 

With this change, users can call `app.CatchSignals(false);` before the main loop to completely skip the installation of these standard signals, leaving them to be handled by the parent process.

### Changes Included
- **`include/ftxui/component/app.hpp`**: Added `void CatchSignals(bool enable = true);` option setter to the `App` interface.
- **`src/ftxui/component/app.cpp`**: Added `bool catch_signals_ = true;` flag to `App::Internal` and conditionalized standard signal installation.
- **`src/ftxui/component/app_test.cpp`**: Added `App.CatchSignals_False` unit test verifying custom signal handlers are preserved and invoked when `CatchSignals(false)` is set.

### Verification Results
- All unit tests compiled and executed successfully.
- Ran `./build/ftxui-tests --gtest_filter="App.*"` (all 15 App-related tests passed).
- Executed the full test suite via `ctest` with a 100% pass rate (331/331 tests passed).

### DIFF
--- a/include/ftxui/component/app.hpp
+++ b/include/ftxui/component/app.hpp
@@ -90,6 +90,13 @@ class FTXUI_EXPORT(COMPONENT) App : public Screen {
   /// @note This feature is only available on POSIX systems (Linux/macOS).
   void HandlePipedInput(bool enable = true);
 
+  /// @brief Enable or disable installing signal handlers for SIGTERM, SIGSEGV,
+  /// SIGINT, SIGILL, SIGABRT, and SIGFPE to restore terminal state on exit.
+  /// @param enable Whether to enable signal handlers. Default is true.
+  /// @note This must be called before Loop().
+  /// @note Signal handlers are enabled by default.
+  void CatchSignals(bool enable = true);
+
   /// @brief Return the currently active app, nullptr if none.
   static App* Active();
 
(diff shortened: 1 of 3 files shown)


ArthurSonzogni:
Thanks!

I understand the problem you are trying to solve, but I'm not sure about the proposed solution.   
  
If we add an option to stop catching standard signals entirely (like `SIGINT` or `SIGSEGV`), any of these signals will cause the application to terminate immediately without cleaning up the terminal state. 
This leaves the terminal in a broken state (raw mode active, cursor hidden, etc.). 
  
We need a way to restore the terminal before propagating or letting the process handle the signal.

Please let me 5 days to resolve this properly. I will request a code review from you.

ArthurSonzogni:
Hello @kghost,

Does this PR fixes your issue?
https://github.com/ArthurSonzogni/FTXUI/pull/1277

kghost:
Thank you very much for your attention.

I have this problem because it conflict with my own crash report, I have not check the solution you have proposed here, I'll check it tomorrow.

ArthurSonzogni:
I am closing this, because I believe this has been fixed now. FTXUI will restore the terminal state in emergency, and raise the original signal handler.

PR #1285, opened 2026-06-03
perf(screen): collapse non-clear ResetPosition cursor walk-up into one CSI

zozowell:
## Problem

In the steady-state redraw path, `Screen::ResetPosition(std::string&, bool)` (the non-`clear` branch) walks the cursor back to the top-left by emitting one `\x1B[1A` (cursor-up) escape **per row**. On a tall terminal this writes a large, redundant burst of escape bytes every single frame. The cursor only needs to be repositioned once — the per-row walk encodes the same movement far more verbosely than necessary.

## The change

Collapse the per-row walk-up into a single parameterized CSI cursor-up, guarded for the single-row case:

```cpp
ss += '\r';  // MOVE_LEFT;
if (dimy_ > 1) {
  ss += "\x1B[" + std::to_string(dimy_ - 1) + "A";  // MOVE_UP;
}
```

The leading `\r` and the final cursor position are unchanged. **On-screen output is identical.**

The `clear` branch is intentionally left per-row: each row needs its own `\x1B[2K` (CLEAR_LINE) erase, so it cannot be collapsed. A code comment notes this.

## Before / after (bytes per frame, non-clear path)

| Rows | Before | After | Reduction |
|-----:|-------:|------:|----------:|
| 10   | 37     | 5     | 7.4x      |
| 24   | 93     | 6     | 15.5x     |
| 50   | 197    | 6     | 32.8x     |
| 100  | 397    | 6     | 66.2x     |

## Tests

Adds `src/ftxui/screen/screen_test.cpp` (registered in `cmake/ftxui_test.cmake`):
- Non-clear output for `dimy_ > 1` equals `"\r\x1B[" + (dimy_-1) + "A"`; equals `"\r"` for `dimy_ == 1`.
- Terminal-effect equivalence to the old per-row walk (same total rows moved up) across several heights — the byte sequences differ by design, so this compares movement, not bytes.
- The 50-row non-clear output is >= 10x smaller than the old per-row form (<= 8 bytes vs 197).
- The `clear` branch remains per-row.

`ctest` passes in full: **339/339 tests pass**, including the new ones.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,11 @@ Next
   opposite direction. Thanks @Ardet696 in #1203.
 
 ### Screen
+- Performance: Collapse the per-row cursor walk-up in the non-clear
+  `Screen::ResetPosition` into a single parameterized CSI cursor-up
+  (`\x1B[<n>A`) instead of emitting one `\x1B[1A` per row. This reduces the
+  per-frame escape bytes during steady-state redraw (e.g. ~197 -> 6 bytes for a
+  50-row screen, ~33x). On-screen output is unchanged.
 - Performance: Optimize `Screen::ToString()`, `Color::Print()` and
   `string_width()`. 
   This was achieved by:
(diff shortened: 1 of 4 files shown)


ArthurSonzogni:
Thanks! That looks useful!

let's wait for the CQ.

PR #1301, opened 2026-06-28
fix: ship Debug and Release libs in Windows package (#1298)

### LINKED ISSUE #1298: /MD and /MDd .lib files
Trying to build FTXUI with /MD instead of /MDd in visual studio gives _ITERATOR_DEBUG_LEVEL mismatches, and CrtDbgReport as an unresolved external. Would it be possible for future releases to include release and debug libs, or is that something we should be building ourselves from source?

ArthurSonzogni:
Closes #1298.

On multi-config generators (Visual Studio), `cmake --install` now places
libraries under `lib/Debug/` and `lib/Release/` instead of a shared `lib/`
directory, so both runtimes (`/MDd` and `/MD`) can coexist in the same
package.

The CI release workflow gains a dedicated Windows step that builds both
configurations and passes `-C "Debug;Release"` to CPack.

A graceful fallback with a `WARNING` is emitted when CMake < 3.20 is
detected (generator expressions in `DESTINATION` require 3.20).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -59,7 +59,8 @@ jobs:
       - name: "Install cmake"
         uses: lukka/get-cmake@latest
 
-      - name: "Build packages"
+      - name: "Build packages (Linux / macOS)"
+        if: runner.os != 'Windows'
         run: >
           mkdir build;
           cd build;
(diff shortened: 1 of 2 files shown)


ArthurSonzogni:
@Hay1tsme - What do you think of this PR? Could you please take a look?

Hay1tsme:
This is just what I was hoping for, thank you! Apologies for the delay.

debashish-ghosh:
This is wrong. My app cannot find the dlls because it's not in the bin directory but rather in bin/Release directory. I use Ninja Multi-Config by default. As a workaround I'm forced to use Ninja generator to have the correct install tree, but this should be fixed in this repo.

PR #1309, opened 2026-07-17
Improve performance of text element.

patlefort:
Hi, this PR bring some improvements to performance for the text element, especially when displaying a large amount of text. It:

* Calculate requirements in constructor since they won't change.
* Draw only visible lines by using the information that it already had.

Example: https://pastebin.com/yGenVLk6

In this example, I can scroll through a million lines no problem with my improvements. Without, it is unusable.

Another possible small improvement would be to change the constructor and the text() functions to use std::string instead of std::string_view, then one could move strings into it, but it would be an ABI break, unless we introduce new text() functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ Changelog
 Next
 ====
 
+### Dom
+- Performance: `text` computes its requirement once and renders only the
+  visible lines. This makes scrolling a large text inside a `frame`
+  significantly faster. Thanks @patlefort. See #1309.
+
 7.0.1 (2026-07-14)
 ------------------
 
--- a/src/ftxui/dom/selection_test.cpp
+++ b/src/ftxui/dom/selection_test.cpp
@@ -152,6 +152,27 @@ TEST(SelectionTest, StyleSelection) {
   }
 }
 
+// The selection style must not persist on an element rendered again after the
+// selection was cleared. See #1309.
+TEST(SelectionTest, StyleClearedWhenSelectionCleared) {
+  auto element = text("Lorem ipsum dolor");
+
+  {
+    auto screen = App::FixedSize(20, 1);
+    Selection selection(2, 0, 9, 0);
+    Render(screen, element.get(), selection);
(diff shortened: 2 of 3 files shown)


ArthurSonzogni:
Thanks, this is a nice improvement. I benchmarked it in Release mode:
- Stock benchmark suite: consistent 2-8% gains, no regressions.
- Your scenario (100k lines scrolled in a 80x24 frame): 15ms → 0.007ms per frame with a reused element (~2000x). Note that reusing an element across frames isn't the supported pattern — FTXUI is designed around rebuilding the DOM from scratch every frame; reuse just happens to work. Rebuilding every frame: 52ms → 39ms; the rest is `Utf8ToGlyphs` and string copies.
- All 347 tests pass. The `check-abi` failure is spurious: fork PRs can't post the report comment. Running the check locally shows no ABI change.
One regression to fix before merging: `has_selection_` and `selection_rows_` were previously reset every frame by `ComputeRequirement()`. Nothing resets them now, since `Select()` is only invoked while a selection is active. With a reused element, the highlight persists after the selection is cleared:
```cpp
auto document = text("Hello");
Screen a(5, 1);
Selection selection(0, 0, 4, 0);
Render(a, document.get(), selection);
Screen b(5, 1);
Render(b, document);  // main: not inverted. This PR: still inverted.
```
Resetting `has_selection_ = false` at the end of `Text::Render()` should be enough: `Select()` runs before `Render()` each frame, so an active selection re-arms it.
Also, please run clang-format on the diff (tabs and `if(` crept in).

ArthurSonzogni:
I will do the "clang-format" and the text selection fix.

patlefort:
I was afraid you might say that pattern isn't supported but it is a necessity in this case. 

It might be possible to create a new text node that take some sort of `ParsedText` struct by reference/pointer that would contain glyphs and lines information. Or maybe a custom component that render a text node with only the visible content. It is hard to know the size of a component because it must be rendered once. Is it a valid pattern in ftxui for a dom element to take something by reference or pointer, outside of other elements?

ArthurSonzogni (approved):

PR #1318, opened 2026-07-30
fix(selection): fill horizontal gaps so mouse-copy keeps spaces

aleroot:
## Mouse text-selection drops inter-word spaces

`Selection::AddPart` only ever concatenates parts; it never reproduces the blank cells in a same-row horizontal gap. The `if (x_ == left + 1)` branch is effectively dead code (identical to the fallthrough), so any layout that separates selectable text with empty columns — notably `paragraph()` / `paragraphAlignLeft`, which splits on spaces and renders words in a `flexbox` with `FlexboxConfig::SetGap(1, 0)` — loses every inter-word space when the selection is copied to the clipboard.

This fills forward same-row gaps with spaces (those cells are inside the selected region and read as spaces on screen), leaving overlapping/out-of-order and cross-row behavior unchanged. Adds a regression test via `paragraph()`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,11 @@ Next
   child, preventing lost focus when entering nested containers whose first
   child is non-focusable. Thanks @Machillka. See #1337.
 
+### Dom
+- Bugfix: Fill horizontal gaps with spaces during text selection so copying text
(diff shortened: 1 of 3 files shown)


ArthurSonzogni:
Thanks!

PR #1319, opened 2026-07-31
Feat: Add gaugeCharset() for gauges with a custom charset

H3X-FF:
### Summary

Adds `gaugeCharset()`, letting a gauge be rendered with a custom set of glyphs instead of the built-in block characters:

```C++
Element gaugeCharset(float progress,
                      std::vector<std::string> charset,
                      Direction direction = Direction::Right);
```

`charset[0]` is the "empty" glyph, `charset.back()` is the "full" glyph; entries in between shade the partially-filled boundary cell. A 2-entry charset gives a plain, unshaded two-glyph bar (e.g. `gaugeCharset(0.5, {" ", "#"})`).

```C++
border(gaugeCharset(0.5, {".", "#"}))
```
```
┌──────────────────────────────────────────────────────────────────────────┐
│#####################################.....................................│
└──────────────────────────────────────────────────────────────────────────┘
```

### History

Original PR by @H3X-FF added this as 12 overloads across `gauge()`/`gaugeLeft()`/`gaugeRight()`/`gaugeUp()`/`gaugeDown()`/`gaugeDirection()`, plus `ftext()`/`fvtext()` (formatted-text helpers via `std::format`). Follow-up commits:

- Collapsed the 12 overloads into the single `gaugeCharset()` above, matching how the rest of `elements.hpp` exposes element variants (`separatorStyled()`, `borderStyled()`: one dedicated name, not overloads of the base function).
- Removed `ftext()`/`fvtext()` — unrelated to the gauge feature, and pulls in `<format>` (C++20) as an unconditional dependency of `elements.hpp`; worth its own separate discussion.
- Reverted a redundant `.gitignore` addition (`.idea/` was already covered by the file's default-deny pattern).
- Added a self-contained doxygen doc comment for `gaugeCharset()` and a `CHANGELOG.md` entry.
- Unified `Gauge::RenderHorizontal`/`RenderVertical`, which had a duplicated near-identical code path for the custom-charset case. That duplication had introduced a real bug: the custom-charset path never called `Invert()`, so `gaugeCharset(..., Direction::Left)` / `Direction::Down` rendered the right glyphs but never flipped cell colors, unlike every other gauge direction. Fixed, with a regression test (`CustomCharsetInvertsLikeDefault`) comparing against `gaugeLeft()`'s inversion behavior.

All tests pass locally (354/354).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@
 # (otherwise causes terribly slow indexing)
 out/
 
+# Ignore CLion's project configs and preferances
+.idea/
+
 # Allowed top-level files:
 !.clang-format
 !.clang-tidy
--- a/.gitignore
+++ b/.gitignore
@@ -7,9 +7,6 @@
 # (otherwise causes terribly slow indexing)
 out/
 
-# Ignore CLion's project configs and preferances
-.idea/
-
 # Allowed top-level files:
 !.clang-format
 !.clang-tidy
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ Next
 - Performance: `text` selection now only visits and stores the selected line
   range, instead of scanning and allocating one entry per line of the whole
   text on every frame.
+- Feature: `gaugeCharset(progress, charset, direction = Direction::Right)`
+  lets a gauge be rendered with a custom set of glyphs instead of the
+  built-in block characters. Thanks @H3X-FF. See #1319.
 
 7.0.1 (2026-07-14)
 ------------------
--- a/include/ftxui/dom/elements.hpp
+++ b/include/ftxui/dom/elements.hpp
@@ -82,12 +82,8 @@ FTXUI_EXPORT(DOM) Element gaugeUp(float progress);
 FTXUI_EXPORT(DOM) Element gaugeDown(float progress);
 FTXUI_EXPORT(DOM) Element gaugeDirection(float progress, Direction direction);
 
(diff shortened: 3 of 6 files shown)


ArthurSonzogni:
Thanks!

I will remove the std::format usage, because it seems unrelated to the PR, and uses C++20 features. We still want to support C++17 users.

ArthurSonzogni:
Thanks for this — pushed a follow-up commit on top of yours with two changes:

**Collapsed the charset overloads into one function.** The 12 overloads across `gauge()`/`gaugeLeft()`/`gaugeRight()`/`gaugeUp()`/`gaugeDown()`/`gaugeDirection()` broke from how the rest of `elements.hpp` exposes variants of an element — `separatorStyled()`, `borderStyled()` use one dedicated name rather than overloading the base function. Also, the 2-char convenience overload was just building a 2-entry vector under the hood, so it didn't need to be a separate overload. Replaced all of it with:

```cpp
Element gaugeCharset(float progress,
                      std::vector<std::string> charset,
                      Direction direction = Direction::Right);
```

`gaugeCharset(0.5, {" ", "#"})` for a plain two-glyph bar, `gaugeCharset(0.5, {" ", "#"}, Direction::Left)` for other directions.

**Dropped `ftext()`/`fvtext()`.** They pulled in `<format>` (C++20) as an unconditional dependency of `elements.hpp` for a formatting convenience unrelated to the gauge feature — worth its own separate PR/discussion rather than bundled here.

Also added a changelog entry. All existing + updated tests pass locally.

ArthurSonzogni (approved):

PR #1322, opened 2026-08-04
cmake: fix exports when modules are enabled.

patlefort:
This fix an issue when c++ modules are enabled. When enabled, cmake export files would be generated without the namespace because the call to `install(EXPORT ...)` in `cmake/ftxui_modules.cmake` would override the previous call. Thus, finding packages `ftxui::*` would fail, except for `ftxui::modules`. The solution: call `install(EXPORT ...)` only once.

I add the `CXX_MODULES_DIRECTORY` property conditionally because it requires a more recent cmake version than the base requirements without modules enabled.


ArthurSonzogni (approved):

ArthurSonzogni:
Thanks!

ArthurSonzogni (approved):

PR #1323, opened 2026-08-04
update actions

ericLemanissier:
mostly update of underlying node version, which fixes the warning about node 20 being deprecated


ArthurSonzogni:
Thanks you!

PR #1332, opened 2026-08-14
Update the Unicode tables from 13.0.0 to 17.0.0

jagerman:
Some characters such the incredibly important U+1FAC3: PREGNANT MAN (🫃) emoji added since the code's last Unicode 13 update (in #484) aren't properly recognized as full-width characters, which breaks how they are drawn.

The PR adds a script to rebuild the full-width and word-break ranges from a given unicode version (for easier future updates), and updates the table to unicode 17.

I've verified (in app) that this fixes the very important pregnant man emoji, as well as a few others in 14-17 updates, added to the unit tests.


ArthurSonzogni (approved):
Excellent! Highly appreciated PR! Thank you!

I moved the generated file to its own file, added a workflow to automated updating the tables, and a CHANGELOG entry.

PR #1333, opened 2026-08-15
Fix Checkbox focus on mouse click

### LINKED ISSUE #959: [bug] Clicking on a Checkbox should give focus to the checkbox.
Clicking on a Checkbox should give focus to the checkbox. It's a particularly unpleasant UX experience, because of the current ambiguity between rendering of focus state and hover state in checkboxes.  

Having TWO checkboxes in apparent focused state is unsettling to begin with (one actually actually focused, and the  other actually in hover state).  It would be mostly forgivable if the focus issue were to resolve itself after a mouse click.  Failing to clear focus on the non-hover Checkbox after a mouse click feels like a serious bug, and it is. Because it happens so often, it violates users' trust, giving them the strong impression that they're dealing with buggy software that can't be relied upon.. 

It doubly sucks for me, because I'm using your library to perform initial setup of a large package, and the very first impression users are going to get is that of broken checkboxes that are controlling  settins for kernel commandline arguments!  If there were ever a situation where violation of trust really counts, that would be it.

A one line fix:

Checkbox.cpp:71

      if (event.mouse().button == Mouse::Left &&
            event.mouse().motion == Mouse::Pressed) {
          *checked = !*checked;
          TakeFocus();   // <--- ADD THIS LINE
          on_change();
          return true;
        }

Not immediately clear whether it should actually be handled in ComponentBase instead. I'll let you be the judge of that. But Button should probably take focus on a mouse click as well (it does not currently). And Radiobuttons. And dropdowns. &c. The advantage of doing it this way is that it's nice to get state changes out of the way before potentially firing any events (in future implementations). 

Note that handling it in a `CheckBoxOptions::on_change` handler is unworkable, since there's no sensible way for the lambda to capture a pointer to a Component that has not yet been created. You might want to think about changing `CheckboxOptions::on_change` from 

      std::function<void()> on_change = [] {};

to 

      std::function<bool(ComponentBase::ptr component)> on_change = [] (ComponentBase::ptr component) { return false; };

if you ever do a major release that has breaking changes.

Machillka:
## Summary

Fixes #959 by making a Checkbox take focus when it is activated with the left mouse button.

## Fix Approach

`TakeFocus()` propagates through the component tree and tells the parent container to make the clicked Checkbox its active child.

The operation order matches the existing keyboard activation path:

1. update the checked state;
2. dispatch the `on_change` callback;
3. take focus.

This keeps mouse and keyboard behavior consistent without introducing new state or modifying other components.

### DIFF
--- a/src/ftxui/component/checkbox.cpp
+++ b/src/ftxui/component/checkbox.cpp
@@ -72,6 +72,7 @@ class CheckboxBase : public ComponentBase, public CheckboxOption {
         event.mouse().motion == Mouse::Pressed) {
       *checked = !*checked;
       App::PostEventOrExecute(on_change);
+      TakeFocus();
       return true;
     }
 


ArthurSonzogni (approved):
Thanks! LGTM!

PR #1335, opened 2026-08-15
Add a cmake uninstall target

debashish-ghosh:


ArthurSonzogni:
Thanks for the contribution!

We generally avoid custom `uninstall` targets in CMake libraries for a few reasons:

1. **Subproject safety:** Reading `@CMAKE_BINARY_DIR@/install_manifest.txt` is unsafe when FTXUI is consumed via `add_subdirectory` or `FetchContent`, as `uninstall` would delete files belonging to the parent project and sibling dependencies.
2. **Fragility:** The install manifest is volatile (lost if `build/` is cleaned, or desynced upon reconfiguration) and leaves empty directory trees behind.
3. **Package management:** Installation lifecycle is better handled via package managers (vcpkg, Conan, Homebrew, system packages) or dedicated isolated `CMAKE_INSTALL_PREFIX` paths (`rm -rf <prefix>`).

For these reasons, we prefer keeping the CMake setup minimal and relying on standard tooling. Closing this, but thanks again for taking the time to put this together!


If anyone believe this should be merged, please +1 and/or start this PR.

PR #1336, opened 2026-08-15
Fix phone number input example accepting line breaks

### LINKED ISSUE #908: "input" example allows line breaks in phone number
Hi there,

I was looking around how to have a number-only input field in ftxui 5.0.0 and came across the `input` example: https://github.com/ArthurSonzogni/FTXUI/blob/5a9ef876a14b91a3cc5b0313cfdaf6b5527eccf4/examples/component/input.cpp#L34-L41

While playing around with this example, I noticed that the "Return" key is allowed to add line breaks to the telephone number. That does not feel like it is intended.

(For my own use I got rid of this by having `multiline = false` in the options of the input component, see https://github.com/ArthurSonzogni/FTXUI/blob/5a9ef876a14b91a3cc5b0313cfdaf6b5527eccf4/src/ftxui/component/input.cpp#L346-L349

Machillka:
## Summary

Fixes #908 by preventing the phone number field in the input example from accepting line breaks.

## Fix Approach

Set `InputOption::multiline` to `false` for the phone number field before creating the `Input` component.

### DIFF
--- a/examples/component/input.cpp
+++ b/examples/component/input.cpp
@@ -32,7 +32,11 @@ int main() {
 
   // The phone number input component:
   // We are using `CatchEvent` to filter out non-digit characters.
-  Component input_phone_number = Input(&phoneNumber, "phone number");
+  InputOption phone_number_option;
+  phone_number_option.multiline = false;
+  Component input_phone_number =
+      Input(&phoneNumber, "phone number", phone_number_option);
+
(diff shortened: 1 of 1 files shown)


ArthurSonzogni (approved):

ArthurSonzogni:
Thanks!

PR #1337, opened 2026-08-16
fix: initialize containers with a focusable child

### LINKED ISSUE #1067: Cursor gets lost with empty sub-container below toggle
Hello,

First of all, awesome library! I've been using it for the last week or so to build a little TUI softphone and it's been great.

I think I found a small UI bug, or I'm doing something wrong. I have a vertical container containing a toggle and another vertical container. This inner vertical container contains an empty horizontal container and a button.

If I launch the project and press the down arrow, my cursor disappears! If I press the down arrow again, it reappears on the button. From there, the UI seems to have "learned" the correct behaviour, and clicking up/down once correctly goes between the toggle and the button.

I threw together a minimum reproducible example: https://gitlab.scd31.com/stephen/ftxui-bug-repro

Here's my main.cpp, if you don't want to click through:

```c++

using namespace ftxui;

int main() {
  auto term = ScreenInteractive::Fullscreen();

  auto btn1 = Button("button 1", [] {});

  auto containerTab1 = Container::Vertical({Container::Horizontal({}), btn1});

  std::vector<std::string> tabValues{
      "Tab1",
  };

  int tabSelected = 0;
  auto tabToggle = Toggle(&tabValues, &tabSelected);

  auto container = Container::Vertical({tabToggle, containerTab1});

  auto renderer = Renderer(container, [&] { return container->Render(); });

  term.Loop(renderer);

  return 0;
}

```

Thanks in advance!

Machillka:
## Summary

Fixes #1067 by ensuring that horizontal and vertical containers initially select a focusable child.

This prevents focus from disappearing when entering a nested container whose first child is not focusable.

## Fix Approach

After initializing the children of a horizontal or vertical container, check whether the selected child is focusable. If not, reuse the existing selector navigation logic to find the next focusable child.

Tab and stacked container behavior remains unchanged.

## Tests

Added a regression test mirroring the structure from the issue: a nested vertical container with an empty first child and a focusable button.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Next
 ### Component
 - Bugfix: Prevent labeled sliders from stretching vertically when placed in a
   container next to taller components. Thanks @Machillka. See #1340.
+- Bugfix: Ensure horizontal and vertical containers initially select a focusable
+  child, preventing lost focus when entering nested containers whose first
+  child is non-focusable. Thanks @Machillka. See #1337.
 
 ### Screen
 - The Unicode tables are updated from 13.0.0 to 17.0.0. Code points assigned by
--- a/src/ftxui/component/container.cpp
+++ b/src/ftxui/component/container.cpp
@@ -69,6 +69,12 @@ class ContainerBase : public ComponentBase {
   int selected_ = 0;
   int* selector_ = nullptr;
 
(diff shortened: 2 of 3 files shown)


ArthurSonzogni (approved):
Thanks!
LGTM!

PR #1342, opened 2026-08-18
fix: propagate active state through CatchEvent

### LINKED ISSUE #1086: CatchEvent() decorator makes component permanently active
Hello!
I have this code to add buttons to a container. Each button has a `CatchEvent` decorator designed to catch keys "a" and "e".
```cpp
for(size_t i = 0; i < m_moduleNames.size(); ++i) {
    Component new_button = 
        Button(m_moduleNames.at(i), [this, i](){   
            if(m_modules.at(i).type != MODULE_EMPTY) {
                m_selectedModule = i;
                m_changeSubwindow(m_subwindowType);
            }
        }, styles::menuStyleButton
    ) | CatchEvent([this, i](Event ev){
        if ( !ev.is_character() || glob.isBlockingGlobalKeys )
            return false;
        
        std::string ch = ev.character();
        if ( ch == "a" || ch == "A") {
            m_appendPM(i);
            return true;
        } else if ( ch == "e" || ch == "E") {
            m_extractPM(i);
            return true;
        }
        return false;            
    });

    m_moduleButtonsContainer->Add(new_button);
}
```

The `CatchEvent` itself works perfectly fine, though it makes all the buttons permanently active. It can be seen since the `styles::menuStyleButton` is:
```cpp
inline ftxui::ButtonOption menuStyleButton {
    .transform = [](const ftxui::EntryState& state) {
        ftxui::Element e;
        if (state.focused) 
            e = ftxui::text(">" + state.label + "<");
        else 
            e = ftxui::text(" " + state.label + " ");
        if (state.focused) 
            e = e | ftxui::bold;
        if (state.active)
            e = e | ftxui::inverted;
        return e;
    }
};
```

When `CatchEvent` is applied, all the components become inverted. 

I tried decorating the main container (`m_moduleButtonsContainer`), but this wouldn't work since i have to get the index of the focused button. Adding `m_moduleButtonsContainer->ActiveChild()->Index()` anywhere in code would break the container entirely -- `ComponentBase::OnRender()` method returns cryptic error message "Not implemented component".

Is this expected behaviour or maybe anyone has a workaround? Thanks in advance!

Machillka:
## Summary

Fixes #1086 by ensuring that `CatchEvent` only exposes its child as active when the wrapper itself is active.

This prevents multiple components wrapped with `CatchEvent` from incorrectly reporting `EntryState.active == true`.

## Fix Approach

Override `CatchEventBase::ActiveChild()` so that it returns the wrapped child only when the `CatchEvent` component is active. Otherwise, it returns `nullptr`.

This preserves the existing event-handling behavior while correctly propagating the container's active state through the wrapper.

## Tests

Added a regression test with two buttons wrapped in `CatchEvent`.

The test verifies that:

- only the selected button is initially active;
- moving the selection updates the active state correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ Next
 - Bugfix: Ensure horizontal and vertical containers initially select a focusable
   child, preventing lost focus when entering nested containers whose first
   child is non-focusable. Thanks @Machillka. See #1337.
+- Bugfix: Propagate the container's active state through `CatchEvent`, preventing
+  inactive wrapped components from incorrectly reporting `Active() == true`.
+  Thanks @Machillka. See #1342.
 
 ### Screen
 - The Unicode tables are updated from 13.0.0 to 17.0.0. Code points assigned by
--- a/src/ftxui/component/catch_event.cpp
+++ b/src/ftxui/component/catch_event.cpp
@@ -17,6 +17,10 @@ class CatchEventBase : public ComponentBase {
       : on_event_(std::move(on_event)) {}
 
   // Component implementation.
+  Component ActiveChild() override {
+    return Active() ? ComponentBase::ActiveChild() : nullptr;
+  }
+
   bool OnEvent(Event event) override {
     if (on_event_(event)) {
       return true;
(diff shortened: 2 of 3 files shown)


ArthurSonzogni (approved):

ArthurSonzogni:
Thanks!

PR #1344, opened 2026-08-21
fix(dom): avoid division by zero on empty table selection

Machillka:
## Summary

Avoid division by zero in `Wrap()` when selecting rows, columns, or rectangles on an empty table (`input_dim_x_ == 0 || input_dim_y_ == 0`).
Also guard `TableSelection::Border()` against empty selections to prevent accessing negative/out-of-bounds indices.

## Tests

Added a unit test `TableTest.EmptySelection` verifying that selections on an empty table execute safely without crashing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ Next
   child, preventing lost focus when entering nested containers whose first
   child is non-focusable. Thanks @Machillka. See #1337.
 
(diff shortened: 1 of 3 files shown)


ArthurSonzogni (approved):

ArthurSonzogni:
Thanks!

PR #1345, opened 2026-08-23
fix: handle consecutive escape sequences

### LINKED ISSUE #675: Bugged character input indentification while mouse event
Like can be seen in [key-press example](https://arthursonzogni.com/FTXUI/examples/?file=component/print_key_press), while holding down ESC "character", and draging the mouse around some random character will spawn instead of the right ones.

Machillka:
## Summary

Fixes #675 by correctly parsing an Escape key immediately followed by a mouse event.

This prevents mouse escape-sequence fragments from being emitted as character input.

## Fix Approach

When a pending Escape byte is immediately followed by another Escape byte, emit the first one as `Event::Escape` and treat the second as the start of the next sequence.

Existing timeout and escape-sequence handling remain unchanged.

## Tests

Added a regression test verifying that an Escape key followed by an SGR mouse report produces one `Event::Escape` and one complete mouse event.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ Next
 ====
 
 ### Component
+- Bugfix: Stop escape sequences from eating the ESC byte starting the next one.
+  Pressing the ESC key while moving the mouse, or any sequence truncated by the
+  terminal, used to be merged with the sequence following it, emitting its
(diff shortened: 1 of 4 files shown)


ArthurSonzogni:
Thanks!

PR #1346, opened 2026-08-27
fix: update selection when a child becomes unfocusable

### LINKED ISSUE #1180: Unfocusable child selected
The
```c++
  int selected_ = 0;
  int* selector_ = nullptr;
```
members of `ContainerBase` should be updated to point to a valid focusable child when some child becomes unfocusable. I had a case where I used `Maybe()` with several `Button()`s in a `Container::Horizontal()` nested in a `Container::Vertical()` and when the button was made unfocusable, when I focused again on the horizontal container, no button was highlighted because `selector_` was not updated.

Machillka:
## Summary

Fixes #1180 by ensuring that horizontal and vertical containers do not keep an unfocusable child selected after its focusability changes dynamically.

## Fix Approach

Revalidate the selected child before rendering horizontal and vertical containers. If it is no longer focusable, reuse the existing selection logic to move to the next focusable child.

## Tests

Added regression tests for horizontal and vertical containers whose initially selected child becomes hidden through `Maybe`.

### DIFF
--- a/src/ftxui/component/container.cpp
+++ b/src/ftxui/component/container.cpp
@@ -108,6 +108,7 @@ class VerticalContainer : public ContainerBase {
   }
 
   Element OnRender() override {
+    EnsureFocusableSelection();
     Elements elements;
     elements.reserve(children().size());
     for (auto& it : children()) {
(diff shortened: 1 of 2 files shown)


ArthurSonzogni:
Thanks!

Conceptually, the render pass isn't supposed to update the state, but I understand why this isn't avoidable in the context of the bug. We can try this PR.

ArthurSonzogni:
Thanks!

PR #1349, opened 2026-09-07
Add ftxui-image-view-async to FTXUI libraries

orrnithogalum:
Adds [ftxui-image-view-async](https://github.com/orrnithogalum/ftxui-image-view-async) to the list of FTXUI libraries.

It is a performance-focused fork of [ftxui-image-view](https://github.com/ljrrjl/ftxui-image-view) that asynchronously loads images and caches the loaded, resized, and converted image data. This avoids repeatedly reading images from disk and converting them to terminal characters on every FTXUI frame redraw, which can otherwise cause significant UI lag in interactive TUIs.

I had to make this because some of my projects fetch images from the internet instead of on-disk, meaning I had to do a network request at every drawn frame which tanked the performance of my application.

Edit: I have also added my projects to the list of public projects in the README.


ArthurSonzogni (approved):
Thanks!

PR #1350, opened 2026-09-13
fix(screen): silence getenv deprecation warning locally

ArthurSonzogni:
The Microsoft CRT marks `std::getenv` as deprecated. CMake only defines `_CRT_SECURE_NO_WARNINGS` for MSVC, so `clang++` targeting Windows reports a `-Wdeprecated-declarations` warning for every call. Reported by @alexv-ds in #900.

## Change

- Add `util::GetEnv(name)` in `src/ftxui/screen/util.hpp`. It suppresses the warning around the single `std::getenv` call only (`#pragma clang diagnostic` for clang, `#pragma warning(disable : 4996)` for MSVC), and returns `""` instead of `nullptr`.
- Use it for the 6 call sites in `terminal.cpp` and `app.cpp`. This replaces the `Safe()` helper and the `safe_getenv` lambda, and drops the `// NOLINT`s.

Why not `#define _CRT_SECURE_NO_WARNINGS` at the top of the file: it only works before the first system header. The amalgamated build (`tools/amalgamate.py`) hoists system headers above every source file, so it would be silently ignored there.

Why not `getenv_s` as in #900: it needs a fixed-size buffer and a UWP special case, while this keeps the behavior identical.

## Testing

- CMake, gcc and clang, `FTXUI_DEV_WARNINGS=ON` (`-Werror`): builds, 366 tests pass.
- Simulated the Microsoft CRT on Linux by declaring `getenv` deprecated before `<cstdlib>` and compiling with clang: 6 warnings before, 0 after. A raw `std::getenv` call still triggers the warning, so the check is meaningful.
- Amalgamated `ftxui.cpp` compiles with gcc and clang under `-Werror`, with 0 simulated warnings.
- Not verified locally: a real clang build on Windows, and the MSVC pragma branch (CMake still defines `_CRT_SECURE_NO_WARNINGS` for MSVC).

Fixes #900.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013jTj3QW3u6qqk7eXF5oWuM


Rucadi:
IDK if it would require too much rewriting, but I feel like std::optional (c++17) (or std::expected (c++23) so better use optional) would be a better return value  than "", otherwise seems like a good PR.

ArthurSonzogni:
Good idea, but that would break backward ABI compatibility.

